PR #1, opened 2021-01-28
FrontRow.App.Http

pbrisbin:
This creates a `FrontRow.App.Http` module tree to (hopefully) serve our needs for making HTTP calls to 3rd parties from any of our backend apps.

## Structure

- `FrontRow.App.Http`: top-level interface, uses/re-exports from below, hopefully serves 95% cases
- `FrontRow.App.Http.Paginate`: conduit-based pagination interface, drop here if you need special behavior (e.g. _not_ via `Link`)
- `FrontRow.App.Http.Retry`: handling of 429-`Retry-In`, unlikely to be independently useful, mostly isolated for organizational tidiness

Modules attempt to be well-documented and rich enough to support all use-cases I found in our backend so far; with the exception of CSV, which, to be honest, I don't remember why I didn't add it at this stage, but we can add it now or later easily.

## Design

The module is `Network.HTTP.Simple`-based, which is itself based on well-accepted `Request` and `Response` types. This should interop with the broader ecosystem nicely when necessary, and be quick to pick up (vs `Wreq`, for example).

Our functions accept a `Request`, but we expose a limited and extended set of request modifiers and builders. Users should not be making `Request -> Request` functions themselves. Some of these could be up-streamed to `.Simple` some day.

Decoding (e.g. JSON) places decoding errors _under_ `Response`. So, `Response (Either String body)`, not `Either e (Response body)`. This, I think, fixes a severe limitation in `.Simple` where you can't look at things like `responseStatus` or `responseHeaders` after you've failed to read the body as JSON.

Concrete use cases for this design popped up in both retries and pagination, where you want to handle the retry headers and paginated requests (which require seeing a `Response`), without short-circuiting to `Left e` on the first JSON decoding error. 

(Decoding is also generalized on `ByteString -> Either e a`, so we support that `httpCsv` easily.)

Pagination is conduit-based, which means it:

1. Supports streaming (of course), but also "just give me them all" trivially, via streaming to `foldC`
2. Composes with any of the `httpX` functions for how to fetch each page

### DIFF
--- a/library/FrontRow/App/Http.hs
+++ b/library/FrontRow/App/Http.hs
@@ -51,6 +51,8 @@ import Data.Bifunctor (first)
 import qualified Data.ByteString as BS
 import Data.ByteString.Lazy (ByteString)
 import qualified Data.ByteString.Lazy.Char8 as BSL8
+import Data.List.NonEmpty (NonEmpty(..))
+import qualified Data.List.NonEmpty as NE
 import FrontRow.App.Http.Paginate
 import FrontRow.App.Http.Retry
 import Network.HTTP.Simple hiding (httpLbs)
@@ -58,35 +60,41 @@ import Network.HTTP.Types.Header (hAccept, hAuthorization)
 import Network.HTTP.Types.Status (Status, statusCode)
 import UnliftIO.Exception (Exception(..), throwIO)
 
-data HttpDecodeError = HttpDecodeError
+data HttpDecodeError e = HttpDecodeError
   { hdeBody :: ByteString
-  , hdeErrors :: String
+  , hdeErrors :: NonEmpty e
   }
   deriving stock (Eq, Show)
 
-instance Exception HttpDecodeError where
+instance Exception e => Exception (HttpDecodeError e) where
   displayException HttpDecodeError {..} =
-    "Error decoding HTTP Response:"
-      <> "\nRaw body: "
-      <> BSL8.unpack hdeBody
-      <> "\nErrors: "
-      <> hdeErrors
+    unlines
+      $ ["Error decoding HTTP Response:", "Raw body:", BSL8.unpack hdeBody]
+      <> fromErrors hdeErrors
+   where
+    fromErrors = \case
+      err :| [] -> ["Error:", displayException err]
+      errs -> "Errors:" : map (bullet . displayException) (NE.toList errs)
+    bullet = (" • " <>)
 
 -- | Request and decode a response as JSON
 httpJson
   :: (MonadIO m, FromJSON a)
   => Request
-  -> m (Response (Either HttpDecodeError a))
-httpJson = httpDecode decode . addAcceptHeader "application/json"
-  where decode body = first (HttpDecodeError body) $ Aeson.eitherDecode body
+  -> m (Response (Either (HttpDecodeError String) a))
+httpJson = httpDecode (first pure . Aeson.eitherDecode)
+  . addAcceptHeader "application/json"
 
 -- | Request and decode a response
 httpDecode
   :: MonadIO m
-  => (ByteString -> Either e a)
+  => (ByteString -> Either (NonEmpty e) a)
   -> Request
-  -> m (Response (Either e a))
-httpDecode decode = fmap (fmap decode) . httpLbs
+  -> m (Response (Either (HttpDecodeError e) a))
+httpDecode decode req = do
+  resp <- httpLbs req
+  let body = getResponseBody resp
+  pure $ first (HttpDecodeError body) . decode <$> resp
 
 -- | Request a lazy 'ByteString', handling 429 retries
 httpLbs :: MonadIO m => Request -> m (Response ByteString)
--- a/library/FrontRow/App/Http.hs
+++ b/library/FrontRow/App/Http.hs
@@ -1,4 +1,4 @@
--- | Centralized module for making HTTP from the backend
+-- | Centralized module for making HTTP requests from the backend
 --
 -- These functions:
 --
--- a/library/FrontRow/App/Http.hs
+++ b/library/FrontRow/App/Http.hs
@@ -7,6 +7,50 @@
 -- - Handle 429-@Retry-In@ for you
 -- - Capture decoding failures with 'Either' values as the 'Response' body
 --
+-- == Examples
+--
+-- Make request, retry on 429s, and parse the body as JSON.
+--
+-- @
+-- -- Throws, but only on a complete failure to perform the request
+-- resp <- 'httpJson' $ 'parseRequest' "https://example.com"
+--
+-- -- Safe access
+-- 'getResponseBody' resp :: Either ('HttpDecodeError' String) a
+--
+-- -- Unsafe access (throws on Left)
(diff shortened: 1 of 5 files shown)


pbrisbin:
I have no idea how to actually test this stuff. Short of launching a small server to talk to during tests. Do we think that's worth it?

library/FrontRow/App/Http.hs
```diff
@@ -0,0 +1,122 @@
+-- | Centralized module for making HTTP from the backend
+--
+-- These functions:
+--
+-- - Do not throw exceptions on non-200
+-- - May throw for other 'HttpException' cases (e.g. 'ConnectionTimeout')
+-- - Handle 429-@Retry-In@ for you
```
eborden, line 7: Do we want retry as the default? This seems like something we may want to avoid when making calls within API handlers, etc.

library/FrontRow/App/Http.hs
```diff
@@ -0,0 +1,122 @@
+-- | Centralized module for making HTTP from the backend
```
eborden, line 1: Some sample code up here might be nice.

library/FrontRow/App/Http.hs
```diff
@@ -0,0 +1,122 @@
...
+--
+-- These functions:
+--
+-- - Do not throw exceptions on non-200
+-- - May throw for other 'HttpException' cases (e.g. 'ConnectionTimeout')
+-- - Handle 429-@Retry-In@ for you
+-- - Capture decoding failures with 'Either' values as the 'Response' body
+--
+module FrontRow.App.Http
```
eborden, line 10: The choice of re-exports is nice.

eborden (approved):
Nothing blocking. Thanks for putting this together.

library/FrontRow/App/Http.hs
pbrisbin: This is a tough trade-off. Every time I tried to make this either opt-in or configurable, it really muddied up the interface, working directly against the primary goal of the module.

I could totally be wrong, but my take is that if we have a Handler that is frequently hitting 429's with a long-enough `Retry-In` that it disrupts our own response-time metrics, would be unlikely to decide that "keep calling that 3rd party synchronously within a request, but just don't retry" is the best answer. I think a simple `forkIO` or moving to a Job are where we'd end up. Therefore, I think this behavior is a feature for that use-case -- it'll make the situation louder.

library/FrontRow/App/Http.hs
pbrisbin: Yeah, good call.

library/FrontRow/App/Http.hs
eborden: That is a reasonable perspective. Just to summarize for my own benefit:

- If retry is taking forever it will inflate the average time of the route.
- We'll notice the route is taking forever.
- We'll likely move the call out to an async process or follow some other mitigation.

Do we have any observability for individual routes?

eborden:
Oh, to answer you original question. Let's do just in time testing for this. The next time it is used we'll just assume experimental status and test then.

library/FrontRow/App/Http.hs
cdparks, line 1: Should this say "making HTTP requests"?

library/FrontRow/App/Http.hs
```diff
@@ -0,0 +1,122 @@
...
+import FrontRow.App.Http.Paginate
+import FrontRow.App.Http.Retry
+import Network.HTTP.Simple hiding (httpLbs)
+import Network.HTTP.Types.Header (hAccept, hAuthorization)
+import Network.HTTP.Types.Status (Status, statusCode)
+import UnliftIO.Exception (Exception(..), throwIO)
+
+data HttpDecodeError = HttpDecodeError
+  { hdeBody :: ByteString
+  , hdeErrors :: String
```
cdparks, line 63: Are there actually multiple errors embedded in one string here?

cdparks (approved):

library/FrontRow/App/Http.hs
pbrisbin: With `Aeson.eitherDecode` that is true, but that's just one use-case.

From the perspective of the library, it's just leaving the function you pass to `httpDecode` in charge of formatting whatever problem or problems it generates into a single message, which I think is a useful separate of responsibility.

library/FrontRow/App/Http.hs
pbrisbin: What, you don't "make HTTP" on the reg?

cdparks: lol not if i can help it

library/FrontRow/App/Http.hs
cdparks: Right, but `httpDecode` doesn't require that the `decode` function return `Either HttpDecodeError a`, only `Either e a`. Since we're only using it for the JSON case, is there a reason to not make this error JSON-specific? Or is the expectation that clients will also want to use `HttpDecodeError`? If the only point of `HttpDecodeError` is to bundle _some_ error with the body, we could do:

```haskell
data HttpDecodeError e = HttpDecodError
  { hdeBody :: ByteString
  , hdeErrors :: NonEmpty e
  }
  deriving (Eq, Show)

instance Exception e => Exception (HttpDecodeError e) where
  displayException HttpDecodeError {..} = intercalate
    "\n"
    [ "Failed to decode HTTP Response"
    , "Raw body: {"
    , BSL8.unpack hdeBody
    , "}"
    , fromErrors hdeErrors
    ]
   where
    fromErrors = \case
      err :| [] -> "Error:\n" <> displayException e
      errs ->
        "Errors:\n" <> intercalate "\n" (bullet . displayException <$> errs)
    bullet = (" • " <>)
```

pbrisbin: I like this, but it's starting to untangle on me. I ended up at:

```hs
-- hdeErrors :: NonEmpty e

httpJson
  :: (..., FromJSON a)
  => ...
  -> m (Response (Either HttpDecodeError String) a)
httpJson = httpDecode $ first pure . Aeson.eitherDecode -- NICE

httpDecode
  :: ...
  => (ByteString -> Either (NonEmpty e) a) -- Hmm, NonEmpty here
  -> Request
  -> m (Response (Either (HttpDecodeError e) a)) -- to e here... OK
```

But, uh,

```hs
instance Exception e => Exception (HttpDecodeError e) where
```

Doesn't work with `e ~ String`.

I thought of,

```hs
instance Show e => Exception (HttpDecodeError e) where
```

But I'll get double-quoting when I `show @String`.

Am I missing some trick or going down a wrong path here?

pbrisbin: Oh whoa, it looks like there's an `Exception` instance for `String`. That's unexpected, but means I can make this work.

library/FrontRow/App/Http.hs
pbrisbin: > Do we have any observability for individual routes?

Yup. Request logging has a `route` field, e.g. `SchoolR`, so we can use CloudWatch Insights to query in complex ways that take that into account ("slowest routes" being on naive example).

PR #4, opened 2021-04-28
Address HttpDecodeError and the lack of Exception(String)

pbrisbin:
**NOTE**: This PR include two options:

- aa2aa0d95edfb47d5cb6b9a4387127a7acffc4dd Remove the e from HttpDecodeErrors

  *This is the option I'm proposing, the end-state of this diff, and so I am
  including the full commit message:*

  `HttpDecodeError String` doesn't have an `Exception` instance, because there
  is no Exception instance for String. Therefore, `getResponseBodyUnsafe` didn't
  work with `httpJson`, which was basically its entire purpose.

  To fix this, we make `HttpDecodeError` hold a `String` internally and ask
  callers to marshal it themselves by passing `ByteString-> (Either (NonEmpty
  String) a` to `httpDecode`.

  Alternatives considered:

  1. Use a `newtype` as necessary in each decoding case

     For example, an `AesonDecodeError` wrapper over `String` with an
     `Exception` instance that just displays it. Similar newtypes would
     effectively be the suggested way to use `httpDecode` with functions that
     use non-`Exception` types in `Left`.

     This alternative was discarded because it is more machinery for no benefit.
     Callers always have to get their decoding error types to a String for us
     somehow. Previously, it was via an `Exception` instance; now, it'd be via
     the decoding function itself. In the previous approach, Exception types
     were handled for free, and non-Exception types would thus require a
     `newtype` and instance.

     In the current approach, String types are handled for free, and other types
     require an extra function call (and `displayException` works great). Given
     that libraries seem to rarely use Exception types for their pure decoders,
     and often use `String`, biasing this way seems like the correct choice.

  2. Use `HttpDecodeError StringException`

     Instead of a dedicated `AesonDecodeError`, we could re-use
     `UnliftIO.Exception.StringException` which serves a similar purpose.

     This alternative was discarded when I saw that `displayException` (via
     `Show`) actually talks about `throwString`, and doesn't just delegate to
     the underlying `String`. No bueno.

- 66541ec70aa8cdf01219ffc5445c9cbd8f8c0962

  This is a commit that tried (1) above first. I kept it in the PR in case you
  would like to see how it turned out concretely and perhaps argue in favor of
  it.

I don't feel strongly between these two options.

### DIFF
--- a/tests/FrontRow/App/HttpSpec.hs
+++ b/tests/FrontRow/App/HttpSpec.hs
@@ -1,0 +1,23 @@
+module FrontRow.App.HttpSpec
+  ( spec
+  ) where
+
+import Prelude
+
+import Control.Lens ((^?))
+import Data.Aeson
+import Data.Aeson.Lens
+import FrontRow.App.Http
+import Network.HTTP.Types.Status (status200)
+import Test.Hspec
+
+spec :: Spec
+spec = do
+  describe "httpJson" $ do
+    it "fetches JSON via HTTP" $ do
+      resp <- httpJson @_ @Value
+        $ parseRequest_ "https://www.stackage.org/lts-17.10"
+
+      getResponseStatus resp `shouldBe` status200
+      body <- getResponseBodyUnsafe resp
+      body ^? key "snapshot" . key "ghc" . _String `shouldBe` Just "8.10.4"
--- a/tests/FrontRow/App/HttpSpec.hs
+++ b/tests/FrontRow/App/HttpSpec.hs
@@ -7,6 +7,8 @@ import Prelude
 import Control.Lens ((^?))
 import Data.Aeson
 import Data.Aeson.Lens
+import Data.Either (isLeft)
+import Data.Text (Text)
 import FrontRow.App.Http
 import Network.HTTP.Types.Status (status200)
 import Test.Hspec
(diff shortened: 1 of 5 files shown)


tests/FrontRow/App/HttpSpec.hs
```diff
@@ -0,0 +1,36 @@
...
+import Data.Text (Text)
+import FrontRow.App.Http
+import Network.HTTP.Types.Status (status200)
+import Test.Hspec
+
+spec :: Spec
+spec = do
+  describe "httpJson" $ do
+    it "fetches JSON via HTTP" $ do
+      resp <- httpJson @_ @Value
```
StevenXL, line 20: Interesting - so this means we let type inference choose the type of `m`, but we are explicitly setting the type of `a`?

StevenXL (approved):

tests/FrontRow/App/HttpSpec.hs
pbrisbin: Yup!

cdparks (approved):
LGTM, I don't remember why we thought there was an `Exception` instances for `String`

pbrisbin:
Rebased out the alternative. Will merge on Green.

PR #5, opened 2021-04-29
Throw HttpStatusError first in getResponseBodyUnsafe

pbrisbin:
Users of `getResponseBodyUnsafe` are typically just looking for a successful
value and to crash otherwise. For unsuccessful responses, the body was unlikely
to parse as they expected, but throwing whatever is the `Left` error hides the
fact that it was unsuccessful to begin with.

This patch will throw a new `HttpStatusError` instead. It still holds the
`Either` response body, so access to all the same information is still present;
the error should just be clearer.

This doesn't change `getResponseBody`, which will return whatever body is there
(`Either` or not) regardless of status. The assumption is that, if you are going
to check the body for `Either` yourself, you can/will also check
`statusIsSuccessful . getResponseStatus` too. For this expected usage, I added
`statusIsSuccessful` to our HTTP re-exports.

### DIFF
--- a/library/FrontRow/App/Http.hs
+++ b/library/FrontRow/App/Http.hs
@@ -78,11 +78,13 @@ module FrontRow.App.Http
   , getResponseBody
 
   -- ** Unsafe access
+  , HttpStatusError(..)
   , getResponseBodyUnsafe
 
   -- * "Network.HTTP.Types" re-exports
   , Status
   , statusCode
+  , statusIsSuccessful
   ) where
 
 import Prelude
@@ -93,15 +95,18 @@ import Data.Aeson (FromJSON)
 import qualified Data.Aeson as Aeson
 import Data.Bifunctor (first)
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.Char8 as BS8
 import Data.ByteString.Lazy (ByteString)
 import qualified Data.ByteString.Lazy.Char8 as BSL8
 import Data.List.NonEmpty (NonEmpty(..))
 import qualified Data.List.NonEmpty as NE
+import Data.Typeable (Typeable)
 import FrontRow.App.Http.Paginate
 import FrontRow.App.Http.Retry
 import Network.HTTP.Simple hiding (httpLbs)
 import Network.HTTP.Types.Header (hAccept, hAuthorization)
-import Network.HTTP.Types.Status (Status, statusCode)
+import Network.HTTP.Types.Status
+  (Status, statusCode, statusIsSuccessful, statusMessage)
 import UnliftIO.Exception (Exception(..), throwIO)
 
 data HttpDecodeError = HttpDecodeError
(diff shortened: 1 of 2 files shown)


library/FrontRow/App/Http.hs
```diff
@@ -168,7 +173,38 @@ addAcceptHeader = addRequestHeader hAccept
...
+  deriving stock Show
+
+instance (Exception e, Typeable a, Show a) => Exception (HttpStatusError e a) where
+  displayException HttpStatusError {..} = unlines
+    [ "Received unsuccessful HTTP Status:"
+    , "Status: " <> show (statusCode hseStatus) <> " " <> BS8.unpack
+      (statusMessage hseStatus)
+    , case getResponseBody hseResponse of
+      Left e -> "Response body did not parse: " <> displayException e
+      Right a -> "Response body: " <> show a
```
StevenXL, line 190: Just thinking out loud here, no changes expected, but if I use `httpJson @_ @Person`, and we get a non-200, it will still try to parse the body to a `Person`, but the body is almost certainly no longer a `Person`, right? We're going to get `Left`s unless the type in the body in the error case can be deserialized to the same type as in the success case. 

Completely minor point, just trying to make sure I am following this properly and not making stuff up.

StevenXL (approved):
LGTM. 

Just thinking out loud here, no changes expected, but if we do get an error, we

library/FrontRow/App/Http.hs
pbrisbin: > but the body is almost certainly no longer a `Person`, right?

Yes that's exactly what would happen. My changes in this PR aren't really meant to change that, but just make the situation downstream of that a _little_ more convenient in _some_ cases.

I'm basing this on the premise that if that's the types you decided to use, that's the only reasonable behavior; and so, all we can really do is try to make it so you see a status-code error first, in cases where it would most likely bite you to not.

Since `getResponseBody` gives you control over the `Either`, you can/should/will be exercising the same control and looking at `getResponseStatus` beforehand, if desired; so I've left that as-is. Since `getResponseBodyUnsafe` is dispatching the `Either` for you, that's where I see the most risk, so that's where I'm injecting the automatic status-check.

But wait, what's the Right Way to address this situation? Well, you shouldn't parse to `Person`!

> unless the type in the body in the error case can be deserialized to the same type as in the success case

Hit the nail right on the head. Example:

```hs
-- However this API structures error messages
data Error = Error
  { code :: Int
  , message :: Text
  }
  deriving stock (Show, Generic)
  deriving anyclass FromJSON

data OrError a = Succeeded a | OrError Error

instance FromJSON a => FromJSON (OrError a) where
  parseJSON v = asum
    [ OrError <$> parseJSON v
    , Success <$> parseJSON v
    ]

main :: IO ()
main = do
  resp <- httpJson @_ @(OrError Person) someRequest

  case getResponseBody of
    Right (Success person) -> doSomethingWithPerson person

    Right (OrError e) | getResponseStatus resp == status400 ->
      logError $ "Bad request: " <> pack (show e)

    Right (OrError e) ->
      logError $ "Other error: " <> pack (show e)

    Left e ->
      logWarn $ "Unexpected response: " <> displayException e
```

I've been trying to come up with a way for `FrontRow.App.Http` to make this pattern easier, but it's so specific to the API in question that I can't.

PR #7, opened 2021-05-04
Revert HttpStatusError, expose setRequestCheckStatus

pbrisbin:
When I tried to use `HttpStatusError` it didn't go well; so I'm pulling that out
and doing something much smaller in scope to see how that goes instead.

Because of the polymorphism of the response body, it was actually impossible to
catch and handle an exception of this type without knowing the concrete response
body we're trying to decode to, which made it impossible for my use-case.

In trying to correct that (which basically means forcing it to `Response ()`), I
found myself literally re-implementing `setRequestCheckStatus`. Therefore, I'm
reverting that other attempt and adding a note to just use that function (which
I'm also re-exporting for that purpose) when that's the behavior you want, which
is what I plan to do where I needed this.

### DIFF
--- a/library/FrontRow/App/Http.hs
+++ b/library/FrontRow/App/Http.hs
@@ -81,9 +81,37 @@ module FrontRow.App.Http
   -- ** Unsafe access
   , getResponseBodyUnsafe
 
+  -- * Exceptions
+  , HttpException(..)
+
+  -- **
+  -- | Predicates useful for handling 'HttpException's
+  --
+  -- For example, given a function 'guarded', which returns 'Just' a given value
+  -- when a predicate holds for it (otherwise 'Nothing'), you can add
+  -- error-handling specific to exceptions caused by 4XX responses:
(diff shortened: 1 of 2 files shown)


package.yaml
```diff
@@ -52,6 +52,7 @@ library:
     - hspec-core
     - hspec-expectations-lifted
     - hspec-junit-formatter
+    - http-client
```
StevenXL, line 55: Non-blocking: If we are using `http-client` simply to get at `HttpException`, is it possible to grab that from `http-conduit`? 

https://www.stackage.org/haddock/lts-17.10/http-conduit-2.3.8/Network-HTTP-Client-Conduit.html#t:HttpException

StevenXL (approved):

package.yaml
pbrisbin: You're right! Will change.

I really dislike how they've sliced up these packages...

PR #9, opened 2021-05-05
Add optional statement timeout to runDB

cbeav:
### Task

[Add statement timeout to Jobs](https://app.asana.com/0/13211253278157/1200281325971161/f)

### Description

We goofed by allowing an application indefinite transaction/statement times. Ideally we could offer configuration for both transaction and statement timeouts here eventually, but the statement level is easy enough to support via PG statement configuration.There is also PG's `idle_in_transaction_session_timeout` which we might be able to leverage if we need further granularity in the future.

### DIFF
--- a/library/FrontRow/App/Database.hs
+++ b/library/FrontRow/App/Database.hs
@@ -10,6 +10,7 @@ module FrontRow.App.Database
   -- * Abstract over access to a sql database
     HasSqlPool(..)
   , SqlPool
+  , Seconds(..)
   , makePostgresPool
   , makePostgresPoolWith
   , runDB
@@ -42,7 +43,7 @@ import System.Process (readProcess)
 
 type SqlPool = Pool SqlBackend
 
-newtype Seconds = Seconds { getSeconds :: Int }
+newtype Seconds = Seconds { unSeconds :: Int }
   deriving stock (Show, Read)
   deriving newtype (Eq, Num)
 
@@ -70,7 +71,7 @@ runDB action = do
 
   flip runSqlPool pool $ do
     for_ mTimeout $ \timeout ->
-      let timeoutMillis = getSeconds timeout * 1000
+      let timeoutMillis = unSeconds timeout * 1000
       in [executeQQ| SET statement_timeout = #{timeoutMillis} |]
     action
 
--- a/library/FrontRow/App/Database.hs
+++ b/library/FrontRow/App/Database.hs
@@ -29,6 +29,7 @@ import Control.Monad.Logger (runNoLoggingT)
 import Control.Monad.Reader
 import Data.ByteString (ByteString)
 import qualified Data.ByteString.Char8 as BS8
+import Data.Foldable (for_)
 import Data.IORef
 import Data.Pool
 import qualified Data.Text as T
@@ -64,7 +65,7 @@ runDB action = do
   mTimeoutSeconds <- asks getStatementTimeoutSeconds
 
   flip runSqlPool pool $ do
-    forM_ mTimeoutSeconds $ \timeoutSeconds ->
+    for_ mTimeoutSeconds $ \timeoutSeconds ->
       let timeoutMilliseconds = timeoutSeconds * 1000
       in [executeQQ| SET statement_timeout = #{timeoutMilliseconds} |]
     action
--- a/library/FrontRow/App/Database.hs
+++ b/library/FrontRow/App/Database.hs
@@ -39,13 +39,10 @@ import Database.Persist.Postgresql
 import Database.Persist.Sql.Raw.QQ (executeQQ)
 import Database.PostgreSQL.Simple (connectPostgreSQL)
 import qualified FrontRow.App.Env as Env
+import FrontRow.App.Time (Seconds(..))
 import System.Process (readProcess)
 
 type SqlPool = Pool SqlBackend
-
-newtype Seconds = Seconds { unSeconds :: Int }
-  deriving stock (Show, Read)
-  deriving newtype (Eq, Num)
 
 class HasSqlPool app where
   getSqlPool :: app -> SqlPool
--- a/library/FrontRow/App/Database.hs
+++ b/library/FrontRow/App/Database.hs
@@ -36,8 +36,9 @@ import Data.Pool
 import qualified Data.Text as T
 import Data.Time (UTCTime, getCurrentTime)
 import Database.Persist.Postgresql
-import Database.Persist.Sql.Raw.QQ (executeQQ)
-import Database.PostgreSQL.Simple (connectPostgreSQL)
+import Database.PostgreSQL.Simple
+  (Connection, Only(..), connectPostgreSQL, execute)
+import Database.PostgreSQL.Simple.SqlQQ (sql)
 import qualified FrontRow.App.Env as Env
 import FrontRow.App.Time (Seconds(..))
 import System.Process (readProcess)
@@ -64,13 +65,7 @@ runDB
   -> m a
 runDB action = do
   pool <- asks getSqlPool
-  mTimeout <- asks getStatementTimeout
-
-  flip runSqlPool pool $ do
-    for_ mTimeout $ \timeout ->
-      let timeoutMillis = unSeconds timeout * 1000
-      in [executeQQ| SET statement_timeout = #{timeoutMillis} |]
-    action
+  runSqlPool action pool
 
 data PostgresConnectionConf = PostgresConnectionConf
   { pccHost :: String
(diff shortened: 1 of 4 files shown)


library/FrontRow/App/Database.hs
```diff
@@ -42,9 +44,11 @@ type SqlPool = Pool SqlBackend
 
 class HasSqlPool app where
   getSqlPool :: app -> SqlPool
+  getStatementTimeoutSeconds :: app -> Maybe Int
```
pbrisbin, line 47: Is this worth a `newtype`?

```suggestion
  getStatementTimeout :: app -> Maybe Seconds
```

library/FrontRow/App/Database.hs
```diff
@@ -58,7 +62,13 @@ runDB
...
-  runSqlPool action pool
+  mTimeoutSeconds <- asks getStatementTimeoutSeconds
+
+  flip runSqlPool pool $ do
+    for_ mTimeoutSeconds $ \timeoutSeconds ->
+      let timeoutMilliseconds = timeoutSeconds * 1000
+      in [executeQQ| SET statement_timeout = #{timeoutMilliseconds} |]
+    action
 
 data PostgresConnectionConf = PostgresConnectionConf
```
pbrisbin, line 73: Should we extend this (and its `env` parser) to look for (something like) `PGSTATEMENT_TIMEOUT`?

I guess it might not be as valuable, since the `PostgresConnectionConf` is typically read, used to make the `Pool SqlBackend` then discarded. Our users would need to stash just part of it on their `App` to use in the instance later. Might still be worth it as a way to standardize the env var.

pbrisbin (approved):

library/FrontRow/App/Database.hs
pbrisbin: I also wonder if we want to actually put a (somewhat high) default here. Having a default was more about not requiring code changes in all users right away; I'm not sure we want "unlimited" to be that default.

library/FrontRow/App/Database.hs
pbrisbin: I take back my hesitance here. I think the common pattern is,

```hs
data AppSettings =
  appPostgresConf :: PostgresConnectionConf

data App =
  appSettings :: AppSettings
```

So this would be useful,

```hs
instance HasSqlPool App where
  -- ...

  getStatementTimeout = pccStatemetnTimeout . appPostgresConf . appSettings
```

library/FrontRow/App/Database.hs
cbeav: Love it. Done and looking good

library/FrontRow/App/Database.hs
```diff
@@ -29,22 +29,30 @@ import Control.Monad.Logger (runNoLoggingT)
...
 import Data.Time (UTCTime, getCurrentTime)
 import Database.Persist.Postgresql
+import Database.Persist.Sql.Raw.QQ (executeQQ)
 import Database.PostgreSQL.Simple (connectPostgreSQL)
 import qualified FrontRow.App.Env as Env
 import System.Process (readProcess)
 
 type SqlPool = Pool SqlBackend
 
+newtype Seconds = Seconds { getSeconds :: Int }
```
pbrisbin, line 45: You'll need to export this, so users can define `getStatementTimeout`. I would probably put it in `FrontRow.App.Seconds` or `FrontRow.App.Time` myself, but exporting it from here seems fine to start too.

pbrisbin: We also have a strong `un`-convention over `get`, for newtypes, FWIW.

```hs
newtype Seconds = Seconds
  { unSeconds :: Int
  }
```

library/FrontRow/App/Database.hs
cbeav: A bit confused. The `PGConf` is only used in `makePostgresPoolWith`, and everything downstream relies on the result type of `SqlPool`. The only `instance` declaration we've got is:
```haskell
instance HasSqlPool SqlPool where
  getSqlPool = id
  getStatementTimeout = const Nothing
```
which has no easy way to anticipate the presence of that conf object?

pbrisbin: Right, that was my initial hesitance

> I guess it might not be as valuable, since the PostgresConnectionConf is typically read, used to make the Pool SqlBackend then discarded

But then I realized the `AppSettings` pattern, which is common.

> The only instance declaration we've got is:

If you look in `jobs`, you'll see the non-`id` usage:

```hs
data App = App
  { appQueue :: Queue
  , appLogLevel :: LogLevel
  -- ...
  , appSqlPool :: SqlPool
  , -- ...
  }

instance HasLogging App where
  getLogLevel = appLogLevel

instance HasSqlPool App where
  getSqlPool = appSqlPool

loadApp :: SqlPool -> IO App
loadApp pool = do
  (..., ..., ...) <- Env.parse $ (,,,,,,,,,,)
    <$> Env.var Env.auto "QUEUE" Env.nonEmpty
    <*> Env.flag (Env.Off LevelInfo) (Env.On LevelDebug) "DEBUG" mempty

  -- ...

  pure $ App
    { appQueue = queue
    , appLogLevel = logLevel
    , appLogFormat = logFormat
    , appSqlPool = pool
    }
```

This is actually a **bad** example:

- It's taking the `pool` into `loadApp` instead of making it there; a conventional `loadApp` should be `:: IO App`
- It's parsing ENV into a massive tuple, ick

This should really be:

```hs
data AppSettings = AppSettings
  { appQueue :: Queue
  , appLogLevel :: LogLevel
  , appPostgresConf :: PostgresConnectionConf -- NEW
  }

instance HasLogging AppSettings where
  getLogLevel = appLogLevel

-- Look ma, no tuple
loadAppSettings :: IO AppSettings
loadAppSettings = Env.parse $ AppSettings
    <$> Env.var Env.auto "QUEUE" Env.nonEmpty
    <*> Env.flag (Env.Off LevelInfo) (Env.On LevelDebug) "DEBUG" mempty
    <*> envParsePostgresConfg -- NEW

data App = App
  { appSettings :: AppSettings
  , appSqlPool :: SqlPool
  }

instance HasLogging App where
  getLogLevel = getLogLevel . appSettings

instance HasSqlPool where
  getSqlPool = appSqlPool
  getStatementTimeout = pccStatementTimeout . appPostgresConf . appSettings -- NEW

loadApp :: IO App
  appSettings@AppSettings {..} <- loadAppSettings
  appSqlPool <- makePostgresPoolWith appPostgresConf  
  pure App { .. }
```

Now, I have a feeling we didn't (or can't) do this because of some complexities with building the Postgres connection using the dynamic IAM credentials. If that's true, fine, we won't refactor Jobs like this. But I still think `PostgresConnectionConfig` should hold the timeout too, and the `env` parser should get it for you, since that makes sense (to me) from the library's perspective.

Not blocking the PR (never was), so feel free to punt on it.

cbeav: Gotcha gotcha. Definitely prefer the pattern of centralizing and preserving `AppSettings`. Happy to try to bend the universe towards that,

library/FrontRow/App/Database.hs
```diff
@@ -58,7 +64,13 @@ runDB
   -> m a
 runDB action = do
   pool <- asks getSqlPool
-  runSqlPool action pool
+  mTimeout <- asks getStatementTimeout
+
+  flip runSqlPool pool $ do
+    for_ mTimeout $ \timeout ->
+      let timeoutMillis = unSeconds timeout * 1000
+      in [executeQQ| SET statement_timeout = #{timeoutMillis} |]
```
jdreaver, line 72: :ghost: 

Doing this in `runDB` adds some round trip latency to every DB transaction. You could consider modifying the pool to set `statement_timeout` when a connection is first created, so all transactions use it implicitly and you don't have to pay for the round trip each time you open a transaction. Alternatively, you could find a way to have persistent send `begin; set statement_timeout = ?;` through the connection instead of just `begin;` when the transaction opens, so you don't have to waste a round trip (since you are sending `begin;` anyway).

Probably not a big deal, but I'm just throwing this out there if you start to see transaction execution times for smaller transactions increase.

pbrisbin: ohai :wave: 

@cbeav this is no different than the implementation in fancy-api right? If the tax David described were to bite us, it would most certainly bite us there.

pbrisbin: Up-streaming an ability to set this on Pool initialization is definitely the best answer though, IMO. I imagine that's `persistent`-owned stuff.

cbeav: Haha, hey there. This is just a complete copy of our current `fancy-api` implementation. We're paying that tax there as well. However, would be nice to pay in neither!

Check out what I just put out. It no longer needs `getStatementTimeout`, but relies on loaded PG config from env. How do we feel about that? I like that it simplifies just relying on the PG config in one place, but it does make this more of a buried feature of the library.

cbeav: Also no tax!

jdreaver: Whoah awesome! I totally forgot we hooked into the pool to do the IAM auth stuff. I thought it would be harder.

PR #10, opened 2021-05-06
Relax use of setRequestIgnoreStatus

pbrisbin:
We decided early on that `FrontRow.App.Http` would not throw exceptions
for non-2XX statuses by default. We did this by using
`setRequestIgnoreStatus` in the base request runner (`httpLbs`).

This was a design choice, but also practically necessary because
`rateLimited` does not function if 429s cause immediate exceptions.

However, the way we did that made it impossible to recover the other
behavior when desired. Any use of `setRequestCheckStatus` would be ignored
because we nope it out deeper down the chain. Users have no recourse
other than explicitly checking status and throwing.

Worse, they can't re-use (or even re-implement!) any of
`Network.Http.Conduit`'s machinery for that check because it is all typed
`Response ReadBody`, which we have lost by the time we've run the request
to `Response body`.

The solution chosen was to remove the blanket ignore, and perform a
smaller modification, specific to `rateLimited` and targeted at 429s. This
is all around better:

- It's less impactful and respects as much user choice as possible
- We still maintain our no-exception design by the fact that we only
  export `parseRequest`/`parseRequest_`, which don't set status-checks

It comes with one minor downside: if a user does chose to check status,
and our `rateLimited` code finds a 429 after its final attempt, that would
not throw, which is surprising. To fix that, we now throw our own
`RetriesExhausted` for this scenario specifically. Unfortunately, we can't
throw a uniform `HttpException` here for the same reasons users who check
status manually can't (no `Response ReadBody`).

### DIFF
--- a/frontrow-app.cabal
+++ b/frontrow-app.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: e421d2d191171413d345575f00769fbfc0c9342b87bbbbbee772e87ceec54c75
+-- hash: e9d517c41cc7d9ff2f36feb6721d9dcbfaec9dd5eebd5ccc48a0baf465f1abb3
 
 name:           frontrow-app
 version:        0.0.0.1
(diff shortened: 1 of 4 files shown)


mjgpy3 (approved):
LGTM!

PR #12, opened 2021-05-06
Fix parsing of optional statement timeout

cbeav:
### Task

Bug fix

### Description

Parsing of this field is broken currently when it is provided, because of weird behavior between `Prelude.read` with `Maybe`. Instead we should try to parse a non-`Maybe` `Seconds` field, and rely on `optional` to do the `Maybe` work for us.

cc @StevenXL who apparently got into a discussion with Pat on Reddit about this once :upside_down_face:

### DIFF
--- a/library/FrontRow/App/Database.hs
+++ b/library/FrontRow/App/Database.hs
@@ -23,7 +23,6 @@ module FrontRow.App.Database
 
 import Prelude
 
-import Control.Applicative (optional)
 import Control.Concurrent
 import qualified Control.Immortal as Immortal
 import Control.Monad.IO.Unlift (MonadUnliftIO)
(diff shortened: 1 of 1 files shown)


pbrisbin (approved):
LGTM. You weren't into the idea of a non-optional just somewhat-high value instead?

cbeav:
> LGTM. You weren't into the idea of a non-optional just somewhat-high value instead?

Oh I'm pretty down with that as well. What's sane, ten minutes?

pbrisbin:
What were you ultimately planning to set `jobs` to once this landed? There's no reason to go any higher than that IMO. Jobs is actually a good fit for "service needing to stretch statement timeouts the most".

cbeav:
> What were you ultimately planning to set `jobs` to once this landed? There's no reason to go any higher than that IMO. Jobs is actually a good fit for "service needing to stretch statement timeouts the most".

Alright at this point I'm just going to require the field and default it to two minutes.

PR #14, opened 2021-06-16
Change withApp to not require a Pool

pbrisbin:
Separating `loadPool` and `withApp` was an unnecessary complexity that arose as
a way to solve us wanting to truncate tables as part of `app` setup in tests.

We thought we would need a handle on the `pool` first to do that, so loading it
once and passing it to the user's `loadApp` was a way to do that.

Two things were discovered:

- Lots of times we don't truncate at all
- We can easily read the pool off a loaded `App` with `HasSqlPool`

So now we just expect `loadApp` to include loading its pool, if necessary. And
we offer a new hook to run something on that pool provided the app `HasSqlPool`.

Users of this library that would previously do:

```hs
main = do
  app <- loadApp =<< makePostgresPool
  -- ...

loadApp pool = App pool
  <*> ...
  <*> ...
```

Can now do

```hs
main = do
  app <- loadApp
  -- ...

loadApp = App
  <$> makePostgresPool
  <*> ...
```

In tests, where one was doing:

```hs
withAppTruncate load = withApp setup
 where
  setup pool = do
    runSqlPool truncateTables pool
    load pool
```

Can now do

```hs
withAppTruncate = withAppSql truncateTables
```

Users of `withApp` would require no changes, once their `loadApp` itself is
updated.

There's going to be a behavior change with regard to when things happen:

**Before**: the `Pool` would be loaded once before all spec items, the
truncation (for example) and loading of `App` would happen before each spec
item.

**After**: the `App` (and its `Pool`) will be loaded once before all spec items,
and (only) the truncation would happen before each spec item.

I think this is an improvement.

### DIFF
--- a/library/FrontRow/App/Test.hs
+++ b/library/FrontRow/App/Test.hs
@@ -22,6 +22,7 @@ import Control.Monad.Random (MonadRandom(..))
 import Control.Monad.Reader
 import Control.Monad.Trans.Control
 import Data.Pool as X
+import Database.Persist.Sql (SqlPersistT, runSqlPool)
 import FrontRow.App.Database as X
 import LoadEnv
 import Test.Hspec as X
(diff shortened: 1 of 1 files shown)


cdparks (approved):
> loading of App would happen before each spec item

In general I think this is fine, but I don't know if we keep any mutable state in our various `App` types, e.g. `IORef`s. Probably not a problem, just a thing to look out for.

pbrisbin:
Yeah, we do some weird stuff in fancy-api for Content, but this won't impact that (pretty sure we're only using this library in Jobs at the moment). And if/when we move it to `FrontRow.App` I'd actually prefer this behavior is as a constraint (since it makes more logical sense, IMO) and rework as necessary to fit whatever API is doing to it.

PR #15, opened 2021-07-01
Bump to latest LTS

pbrisbin:
This updated http-link-header with a changed interface. I had a .Compat module
from when yesod-page-cursor went through this, so I just copied it in here. We
could consider releasing http-link-header-compat, as is a common pattern for
this, but I'm going to leave the copy/paste for now.


library/Network/HTTP/Link/Compat.hs
```diff
@@ -0,0 +1,27 @@
...
+where
+
+import Prelude
+
+import Data.Text (Text)
+import Network.HTTP.Link hiding (Link)
+import qualified Network.HTTP.Link as HTTP
+import Network.URI (URI)
+
+#if MIN_VERSION_http_link_header(1,2,0)
```
mjgpy3, line 17: I think I remember seeing some discussion about this :grimacing:

mjgpy3 (approved):
Looks good to me. I appreciate the comment about tests creating commits.

PR #16, opened 2021-07-09
Add an export for setRequestBodyURLEncoded

z0isch:
https://app.asana.com/0/1199344717368182/1200421645749559/f is requiring setting some `x-www-form-urlencoded` `POST` request body.

This exposes the function `Network.HTTP.Simple.setRequestBodyURLEncoded` to do so


pbrisbin (approved):
Only nit might be to move to adjacent to the other `setRequestBody` function in the export list.

PR #24, opened 2021-08-19
Separate test suite.

StevenXL:
The `VersionSpec` specs have been extracted from the `spec` test suite /
target into a `gittest` target.

The `gittest` target will only be built if the `git` flag is passed.
This is the **default**, but it allows us to ignore this test suite in
specific circumstances.

The reason for this split is that [we can't run
git](https://github.com/commercialhaskell/stackage/issues/6172) when
stackage is running our test suite.


StevenXL:
@pbrisbin I know that we could have used something a big more general in the flag name and the target name, but I figure that if the time comes to do that, it will be easy enough.

freckle-app.cabal
```diff
@@ -23,6 +23,11 @@ source-repository head
   type: git
   location: https://github.com/freckle/freckle-app
 
+flag git
```
pbrisbin, line 26: I might call this `test-git` or something? A flag just named `git` might imply you are disabling all git functionality, not just the testing thereof.

pbrisbin (approved):
Neat fix!

For what it's worth we frequently tell Stackage to just not care about our suite failing (see yesod-page-cursor, for example, which requires a DB to run tests), but this is neat too.

PR #25, opened 2021-08-20
Extract tests that require `git` into new suite.

StevenXL:
New release. This will allow us to add `freckle-app` on stackage without having to ignore the test suite (hopefully).


StevenXL:
@pbrisbin The GitHub review round-robin seems to keep picking you. (No coincidence since it is looking at whose worked on the repo).

pbrisbin (approved):

PR #31, opened 2021-10-05
Add function for responding 503 on query-cancelled

pbrisbin:
When the `statement_timeout` set by `Freckle.App.Database` triggers, we get this
`SqlError`. By attaching one of these functions as `yesodMiddleware`, we can
turn that into a 503.

The function also logs, increments a DogStats metric, and supports attaching
headers (necessary for our AuthorizationChecked machinery).


StevenXL (approved):
LGTM!

PR #34, opened 2021-11-02
Enable reporting RTS stats to datadog

eborden:
This capability existed as a one off within a single service. This
information can be useful to determine whether RTS behavior is leading
to CPU or memory issues and is simple to abstract.

This leans on `ekg-core` to collect the stats about the RTS. It then
initializes a loop to poll that information and send it to DataDog.


pbrisbin (approved):
Sick, so you're doing https://app.asana.com/0/0/1201218947043799/f for me?

eborden:
Yerp

PR #37, opened 2021-11-05
Add flow URL code generation

mjgpy3:


mjgpy3:
Dropping since we got maintainer rights on the package.

PR #39, opened 2021-11-09
Use Template Haskell to produce routes names

eborden:
Yesod handles dispatch through abstract types and template haskell.
These opaque `Route a` types makes it hard to traverse the data without
manually unrolling the entire tree yourself. Instead we can use the same
methods that `yesod` uses internally to traverse `ResourceTree` and
produce the function from the route file itself.

This template haskell is pretty tame, it just builds a gigantic lambda
case expression nesting all of the routes and returning the leaf route
name at the end.

Other solutions are possible since we hold onto all `Static` and
`Dynamic` path names in the route file, but this method matches our
current use of `____R` resource names.

### DIFF
--- a/library/Freckle/App/Yesod/Route.hs
+++ b/library/Freckle/App/Yesod/Route.hs
@@ -1,0 +1,63 @@
+module Freckle.App.Yesod.Route
+  ( mkRouteNameCaseExp
+  ) where
+
+import Prelude
+
+import Data.Foldable (fold)
+import qualified Language.Haskell.TH as TH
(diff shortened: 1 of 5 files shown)


pbrisbin (approved):

library/Freckle/App/Yesod/Route.hs
```diff
@@ -0,0 +1,64 @@
+module Freckle.App.Yesod.Route
```
eborden, line 1: I'm going to change this to `Routes` to match yesod's structure

PR #41, opened 2021-12-03
Add middleware for denying all HTML frame embedding

mjgpy3:
From the task

## Description
The "X-Frame-Options" header controls which origins are allowed to iframe the current page. The lack of this header can make many attacks such as XSS, CSRF, and Client-Side Account Takeover much more easily exploitable. Without the "X- Frame-Options" header it is also possible to perform click-jacking attacks which may have high impact depending on the application.

## Remediation
It is recommended that the "X-Frame-Options" be added to the response headers throughout the application in order to prevent easy exploitation of client-side bugs and click-jacking.


StevenXL (approved):
LGTM.

I don't think we need the `assertBody` but doesn't hurt.

PR #42, opened 2021-12-07
Update to hspec-unit-formatter-1.1.0.0

pbrisbin:
This version removed `runJUnitSpec` in favor of functions better-suited to the
common case (e.g. `hspecJUnit`). This, however, is not the common case.

We are running three suites separately and evaluating everything once at the
end. Therefore, we have to dig around in hspec internals a little more and
re-implement `runJUNitSpec` ourselves.

### DIFF
--- a/library/Freckle/App/Test/Hspec/Runner.hs
+++ b/library/Freckle/App/Test/Hspec/Runner.hs
@@ -3,8 +3,7 @@ module Freckle.App.Test.Hspec.Runner
   , runParConfig
   , runWith
   , makeParallelConfig
-  )
-where
+  ) where
 
 import Prelude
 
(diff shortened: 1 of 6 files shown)


pbrisbin:
My commit message is wrong, it's 1.1.0.0 not 1.0.0.0. Will amend.

library/Freckle/App/Test/Hspec/Runner.hs
```diff
@@ -14,12 +13,15 @@ import Control.Monad (void)
 import Control.Monad.Trans.Maybe (MaybeT(MaybeT), runMaybeT)
 import Data.List (isInfixOf)
 import Data.Maybe (fromMaybe, isJust)
+import Data.Text (pack)
 import System.Environment (getArgs, lookupEnv)
-import Test.HSpec.JUnit (runJUnitSpec)
 import Test.Hspec (Spec)
+import Test.Hspec.JUnit
+  (configWithJUnit, defaultJUnitConfig, setJUnitConfigOutputFile)
```
pbrisbin, line 20: Actually this works with `v1.0.3.0` so I think we can avoid a major version bump ourselves. Let me try things without the exra deps.

pbrisbin: Yup works fine.

stack.yaml
```diff
@@ -1,4 +1,10 @@
-resolver: lts-18.2
+resolver: lts-18.18
+
+extra-deps:
+  - hspec-2.9.3
+  - hspec-core-2.9.3
+  - hspec-discover-2.9.3
+  - hspec-junit-formatter-1.1.0.0
```
z0isch, line 7: Does this be `hspec-junit-formatter-1.0.3.0` then?

pbrisbin: I tested it with that, but decided to still use latest for local development.

pbrisbin: Similarly, we work fine with hspec-2.8, but I added 2.9 here anyway.

z0isch (approved):

PR #44, opened 2022-01-10
Add Freckle-specific prelude

mjgpy3:
**Why?**

In removing `classy-prelude` from our monorepo, a ton of import ergonomics have
been lost. This attempts to recover some of those ergonomics and also be the
starting point for our very own prelude.

This was mostly inspired by the things I had to import a gazillion types when
removing `classy-prelude` from a few of our libraries. Before hitting the
library that has the most instances, I thought it'd be nice to have something
like this in place.

**Things to look out for**

 - Did I export too much?
 - Are there glaring omissions?
 - Did I pick the wrong thing in some case (e.g. is `Control.Monad.Reader` the
 - "right" reader?)
 - Should this move into our monorepo?
 - Should I get a dep from another package? All new packages were hidden deps, which is nice, but should I have instead just squashed RIO + Prelude?
  - Do _YOU_ specifically have a prelude preference that we should consider instead?
  - Should this be in freckle-app or elsewhere?

**Note on reviewing**

This is likely controversial so I'll be sure to eyes from all interested
backenders.

### DIFF
--- a/library/Freckle/App/Database.hs
+++ b/library/Freckle/App/Database.hs
@@ -22,11 +22,10 @@ module Freckle.App.Database
   , envPostgresPasswordSource
   ) where
 
-import Prelude
+import Freckle.App.Prelude
 
 import Control.Concurrent
 import qualified Control.Immortal as Immortal
-import Control.Monad.IO.Unlift (MonadUnliftIO)
 import Control.Monad.Logger (runNoLoggingT)
 import Control.Monad.Reader
 import Data.ByteString (ByteString)
@@ -35,12 +34,12 @@ import Data.Char (isDigit)
 import Data.IORef
 import Data.Pool
 import qualified Data.Text as T
-import Data.Time (UTCTime, getCurrentTime)
 import Database.Persist.Postgresql
 import Database.PostgreSQL.Simple
   (Connection, Only(..), connectPostgreSQL, execute)
 import Database.PostgreSQL.Simple.SqlQQ (sql)
 import qualified Freckle.App.Env as Env
+import qualified Prelude as Unsafe (read)
 import System.Process (readProcess)
 
 type SqlPool = Pool SqlBackend
@@ -117,9 +116,10 @@ readPostgresStatementTimeout
   :: String -> Either String PostgresStatementTimeout
 readPostgresStatementTimeout x = case span isDigit x of
   ("", _) -> Left "must be {digits}(s|ms)"
-  (digits, "") -> Right $ PostgresStatementTimeoutSeconds $ read digits
-  (digits, "s") -> Right $ PostgresStatementTimeoutSeconds $ read digits
-  (digits, "ms") -> Right $ PostgresStatementTimeoutMilliseconds $ read digits
+  (digits, "") -> Right $ PostgresStatementTimeoutSeconds $ Unsafe.read digits
+  (digits, "s") -> Right $ PostgresStatementTimeoutSeconds $ Unsafe.read digits
+  (digits, "ms") ->
+    Right $ PostgresStatementTimeoutMilliseconds $ Unsafe.read digits
   _ -> Left "must be {digits}(s|ms)"
 
 envPostgresPasswordSource :: Env.Parser PostgresPasswordSource
--- a/library/Freckle/App/Prelude.hs
+++ b/library/Freckle/App/Prelude.hs
@@ -1,0 +1,147 @@
+-- | Those functions and types we can't do without!
+module Freckle.App.Prelude
+  (
+  -- * 'Prelude' as the starting point, without common partial functions
+    module Prelude
+
+  -- * Commonly imported types
+  , Alternative
+  , Exception
+  , Generic
+  , HashMap
+  , HashSet
+  , Hashable
+  , Int64
+  , Map
+  , MonadIO
+  , MonadReader
+  , MonadUnliftIO
+  , NominalDiffTime
+  , NonEmpty
+  , PrimMonad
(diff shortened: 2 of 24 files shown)


stackptr:
Going to follow the conversation on this PR but removing my review to defer to those with more opinions about the prelude

library/Freckle/App/Database.hs
```diff
@@ -22,11 +22,11 @@ module Freckle.App.Database
   , envPostgresPasswordSource
   ) where
 
-import Prelude
+import Freckle.App.Prelude
+import Prelude (read)
```
pbrisbin, line 26: When doing something like this (using a safe prelude and an unsafe function from `Prelude`). My convention is to `import Prelude as Unsafe (read)`. Then it's nice and clear that you're using `Unsafe.read` later. I also don't consider the `Unsafe` import a "prelude", so I group it down with the rest.

library/Freckle/App/Prelude.hs
```diff
@@ -0,0 +1,136 @@
...
+  , (<=<)
+  , (>=>)
+  , guard
+  , join
+  , unless
+  , void
+  , when
+
+  -- ** 'Arrow'
+  , (&&&)
```
pbrisbin, line 82: If you've got this, let's add `(***)` too?

pbrisbin (approved):
I don't feel strongly what exactly this looks like; I think we all just wanted to start conservative and expand in response to concrete need; so this LGTM!

I just wonder: just how close did you land to re-creating `Relude` or `RIO.Prelude` here?

library/Freckle/App/Database.hs
mjgpy3: I like that -- it makes things pretty clear.

mjgpy3:
> I just wonder: just how close did you land to re-creating Relude or RIO.Prelude here?

That's a good question, a preliminary glance reveals that this is pretty close.

library/Freckle/App/Prelude.hs
```diff
@@ -0,0 +1,137 @@
...
+  , on
+
+  -- ** 'UTCTime'
+  , getCurrentTime
+  ) where
+
+-- Use 'Prelude' as the starting point, removing common partial functions
+
+import Prelude hiding
+  (cycle, foldl1, foldr1, head, init, last, maximum, minimum, read, tail, (!!))
```
eborden, line 95: I love that you are hiding some partial functions. Can we export their friends from `Safe` (`headMay`, `readMay`, etc) and `semigroupoids` (`fold1`, `foldMap1`, etc).

library/Freckle/App/Prelude.hs
```diff
@@ -0,0 +1,137 @@
...
+module Freckle.App.Prelude
+  (
+  -- * 'Prelude' as the starting point, without common partial functions
+    module Prelude
+
+  -- * Commonly imported types
+  , Alternative
+  , Exception
+  , Generic
+  , HashMap
```
eborden, line 11:
```suggestion
  , HashMap
  , HashSet
```

library/Freckle/App/Prelude.hs
```diff
@@ -0,0 +1,137 @@
...
+  , isNothing
+  , listToMaybe
+  , mapMaybe
+  , maybeToList
+  , readMay
+
+  -- ** 'Either'
+  , partitionEithers
+
+  -- ** 'Foldable'
```
eborden, line 49: Honestly, I'd just export all of `Foldable` and `Traversable`, maybe with the `_1` functions hidden, in favor of their `semigroupoid` version.

library/Freckle/App/Prelude.hs
```diff
@@ -0,0 +1,137 @@
...
+import Data.Function (on)
+import Data.Maybe
+  (catMaybes, fromMaybe, isJust, isNothing, listToMaybe, mapMaybe, maybeToList)
+import qualified Data.Text as T
+import Data.Traversable (for)
+import Safe (readMay)
+
+infixl 4 <$$>
+(<$$>) :: (Functor c, Functor d) => (a -> b) -> c (d a) -> c (d b)
+f <$$> as = (f <$>) <$> as
```
eborden, line 133: We have a bunch of modules in `megarepo` like `FrontRow.Core.Applicative` with functions like this that are probably worth migrating here. I don't want that work to block this, but it could be a good followup task.

eborden (approved):
This looks really solid and matches what we already do, just removing a lot of the pain. I have nothing blocking, just a few thoughts.

We probably want to consider coming up with rules for what gets included in a prelude. Those decision points will arise and we should make it easier on ourselves by having a rubric (maybe with a comment action when this file is changed.)

library/Freckle/App/Prelude.hs
mjgpy3: Sounds good but aren't the `_{n}` functions from `Control.Lens`?

StevenXL (approved):
Really appreciate that the Freckle Prelude has evolved from concrete problems you ran into. LGTM!

library/Freckle/App/Prelude.hs
eborden: Oh sorry, I meant things like `foldl1`, not `_1` and `_2`. Maybe `____1` would have been more understandable?

mjgpy3: Oh, gotcha!

mjgpy3:
I think I'm going to remove `on`. Looking into our code a little more we hide it constantly in favor of `Database.Esqueleto (on)`.

PR #45, opened 2022-01-12
Remove explicit region specification in DB token generation

cbeav:
This hard-coded region prevents deploys of `freckle-app` backed by IAM auth outside of `us-east-1`. We can remove it and allow for region discovery in AWS environments to populate as desired (assuming co-location of the deployment with the target DB).


pbrisbin (approved):

pbrisbin:
Going to merge this because I'm cutting a new release soon.

PR #47, opened 2022-02-07
Test on multiple GHCs, release with looser lower-bounds

pbrisbin:
Introduce a matrix and test on multiple older GHCs on CI.

Use the oldest resolver we test with on release. This ensures the lowest lower-bounds we can in the published package, and should address the Haddocks failures on Hackage.


library/Freckle/App/Test.hs
```diff
@@ -15,6 +15,7 @@ import Freckle.App.Prelude
 
 import Control.Monad.Base
 import Control.Monad.Catch
+import qualified Control.Monad.Fail as Fail
```
pbrisbin, line 18: This is the recommend approach, compability-wise. The import is necessary on older GHCs, and redundant on newer ones. Doing it qualified allows it to be present for either.

library/Freckle/App/Database.hs
```diff
@@ -35,6 +35,13 @@ import Data.IORef
 import Data.Pool
 import qualified Data.Text as T
 import Database.Persist.Postgresql
+  ( SqlBackend
+  , SqlPersistT
+  , createPostgresqlPoolModified
+  , createSqlPool
+  , openSimpleConn
+  , runSqlPool
+  )
```
pbrisbin, line 44: Some versions of this module export a `Connection`, others don't. Explicit imports allow us to avoid the collision with `PostgreSQL.Simple` in the versions that do.

library/Freckle/App/Wai.hs
```diff
@@ -159,7 +159,7 @@ readForwardedFor bs
 sockAddrToIp :: SockAddr -> ByteString
 sockAddrToIp (SockAddrInet _ h) = BS8.pack $ show $ fromHostAddress h
 sockAddrToIp (SockAddrInet6 _ _ h _) = BS8.pack $ show $ fromHostAddress6 h
-sockAddrToIp (SockAddrUnix _) = "<socket>"
+sockAddrToIp _ = "<socket>"
```
pbrisbin, line 162: Older versions of `network` have an additional constructor, eventually deprecated then removed. Though we don't like the `_`, it's a convenient way to paper over that.

package.yaml
```diff
@@ -90,11 +90,11 @@ library:
     - primitive
     - process
     - resource-pool
-    - retry
+    - retry >= 0.8.1.0
```
pbrisbin, line 93: We use `retryingDynamic`, introduced in this version. I discovered it because it's not present in `lts-12`.

freckle-app.cabal
```diff
@@ -115,7 +115,7 @@ library
     , lens
     , load-env
     , monad-control
-    , monad-logger
+    , monad-logger >=0.3.31
```
pbrisbin, line 118: Adds the `fromLogStr` re-export we use.

freckle-app.cabal
```diff
@@ -102,20 +102,20 @@ library
...
-    , hspec
-    , hspec-core
+    , hspec >=2.8.1
+    , hspec-core >=2.8.1
     , hspec-expectations-lifted
-    , hspec-junit-formatter
+    , hspec-junit-formatter >=1.1.0.1
     , http-client
-    , http-conduit
+    , http-conduit >=2.3.5
```
pbrisbin, line 110: `addToRequestQueryString`

.github/workflows/release.yml
```diff
@@ -21,3 +21,6 @@ jobs:
           pvp-bounds: lower
         env:
           HACKAGE_API_KEY: ${{ secrets.HACKAGE_UPLOAD_API_KEY }}
+
+          # The oldest LTS we CI against, to set the lowest lower-bounds we can
+          STACK_YAML: stack-lts-12.26.yaml
```
StevenXL, line 26: Do you think it makes sense to add a comment to the matrix stanza in the `ci` workflow? If someone where to delete `12.26` from the matrix, they might not know to come over to the `ci` workflow and update this.

StevenXL (approved):

.github/workflows/release.yml
pbrisbin: That's a good note. I was thinking of ways to make it unnecessary.

For example, `stack-upload-action` itself could check for `stack-lts-*.yaml` and use the lowest one automatically.

This would be easy to do and most convenient for us, but it's a bit opinionated, so I'm not sure the best way to do it options-wise for general consumption.

PR #48, opened 2022-02-18
Add some common textual encoding functions to prelude

mjgpy3:
These were discussed in backend guild as being common enough that we should
include them in `Freckle.App.Prelude`.

### DIFF
--- a/library/Freckle/App/Prelude.hs
+++ b/library/Freckle/App/Prelude.hs
@@ -33,6 +33,10 @@ module Freckle.App.Prelude
 
   -- ** 'Text'
   , tshow
+  , pack
+  , unpack
+  , encodeUtf8
+  , decodeUtf8
 
   -- ** 'Maybe'
   , catMaybes
@@ -103,7 +107,7 @@ import Data.Int (Int64)
 import Data.List.NonEmpty (NonEmpty)
 import Data.Map.Strict (Map)
 import Data.Set (Set)
-import Data.Text (Text)
+import Data.Text (Text, pack, unpack)
 import Data.Time (NominalDiffTime, UTCTime, getCurrentTime)
 import Data.Vector (Vector)
 import GHC.Generics (Generic)
@@ -135,7 +139,7 @@ import Data.Either (partitionEithers)
 import Data.Foldable as Foldable hiding (foldl1, foldr1)
 import Data.Maybe
   (catMaybes, fromMaybe, isJust, isNothing, listToMaybe, mapMaybe, maybeToList)
-import qualified Data.Text as T
+import Data.Text.Encoding (decodeUtf8, encodeUtf8)
 import Data.Traversable as Traversable
 
 infixl 4 <$$>
@@ -144,4 +148,4 @@ f <$$> as = (f <$>) <$> as
 
 -- | 'Show' as 'Text'
 tshow :: Show a => a -> Text
-tshow = T.pack . show
+tshow = pack . show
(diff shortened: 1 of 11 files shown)


eborden (approved):
One more hop to `Freckle.Prelude` :laughing:

eborden (commented):
Do you have a sense of the ubiquity? I can guess it, but I'd be curious to know how many imports this is going to wipe out.

library/Freckle/App/Wai.hs
```diff
@@ -11,7 +11,7 @@ module Freckle.App.Wai
   , denyFrameEmbeddingMiddleware
   ) where
 
-import Freckle.App.Prelude
+import Freckle.App.Prelude hiding (decodeUtf8)
```
pbrisbin, line 14: What's conflicting?

library/Freckle/App/Prelude.hs
```diff
@@ -135,7 +139,7 @@ import Data.Either (partitionEithers)
 import Data.Foldable as Foldable hiding (foldl1, foldr1)
 import Data.Maybe
   (catMaybes, fromMaybe, isJust, isNothing, listToMaybe, mapMaybe, maybeToList)
-import qualified Data.Text as T
+import Data.Text.Encoding (decodeUtf8, encodeUtf8)
```
pbrisbin, line 142: I might've mentioned it in the discussion, but for the record:

`RIO` and any others that follow the "no unsafe functions by default" philosophy don't export this, since it's technically unsafe / lossy. They doe export `decodeWith` and `lenientDecode` so you can make it easily-but-explicitly yourself.

I'm conflicted on proposing we do that -- One of the very first thing I add to my `${App}.Prelude` in an `RIO` project is define a `decodeUtf8`, but figured I'd raise it.

Separate form that, what about _also_ adding `decodeUtf8With` and the various `.Errors` options? The `Wait` module in this project itself is one example that would benefit.

pbrisbin (approved):
@eborden roughly,

```console
% cd megarepo/backend
% g grep -E '(T\.)?(un)?pack\b' | wc -l
468
% g grep -E '(T\.)?(de|en)?codeUtf8(With)?\b' | wc -l
181
```

And the number of unique files, presumably how many actual `import` lines would be dropped by this:

```console
% g grep -E '\b(T\.)?(un)?pack\b' | cut -d: -f1 | uniq | wc -l
178
% g grep -E '\b(T\.)?(en|de)?codeUtf8\b' | cut -d: -f1 | uniq | wc -l
70
```

library/Freckle/App/Wai.hs
mjgpy3: We have the following defined below
```haskell
decodeUtf8 :: ByteString -> Text
decodeUtf8 = decodeUtf8With lenientDecode
```

pbrisbin: Isn't that what `Prelude.decodeUtf8` is? In other word, why hide instead of removing that?

mjgpy3: Looking at `Data.Text.Encoding` it looks like it's defined as
```haskell
decodeUtf8 = decodeWithUtf8 strictDecode
```

Have I exported the wrong thing?

pbrisbin: Good question. I didn't realize that and have always operated under the assumption it was `lenientDecode`. Probably a decent argument for not exporting this and requiring explicit local redefinition :upside_down_face: .

If everyone discussing this in Guild knew that but me then sounds like this PR is totally right as-is. If others were confused or didn't think about this, it could be worth re-opening the discussion. :shrug: 

For what it's worth, when I wrote the local `decodeUtf8` here, I did not pick `lenientDecode` on purpose, I was simply attempting to redefine the normal one. So no `hiding` and removing it would change behavior but match original intent.

PR #53, opened 2022-05-18
Freckle.App.Memcached

pbrisbin:
- Freckle.App.Memcached
- Add Memcached on CI

### DIFF
--- a/library/Freckle/App/Memcached/Servers.hs
+++ b/library/Freckle/App/Memcached/Servers.hs
@@ -8,14 +8,25 @@
 -- memcached://[user[:password]@]host][:port],...
 -- @
 --
+-- Usage with "Freckle.App.Env":
+--
+-- @
+-- -- Required
+-- Env.var (Env.eitherReader readMemcachedServers) "MEMCACHED_SERVERS" Env.nonEmpty
+--
+-- -- Default to localhost:11211
+-- Env.var (Env.eitherReader readMemcachedServers) "MEMCACHED_SERVERS" (Env.def defaultMemcachedServers)
+--
+-- -- Default to disabled
+-- Env.var (Env.eitherReader readMemcachedServers) "MEMCACHED_SERVERS" (Env.def emptyMemcachedServers)
+-- @
+--
 module Freckle.App.Memcached.Servers
   ( MemcachedServers(..)
   , defaultMemcachedServers
-  , envParseMemcacheServers
-  , toServerSpecs
-
-  -- * Exported for testing
+  , emptyMemcachedServers
   , readMemcachedServers
+  , toServerSpecs
   ) where
 
 import Freckle.App.Prelude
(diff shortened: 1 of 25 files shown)


pbrisbin:
Once this is good to go, I'll make the `nightly` status not required and `stack-lts-nightly.yaml` required instead.

pbrisbin:
@eborden you reviewed most of this over in the megarepo PR. There are some minor changes in the details, but it's mostly just a shift-and-lift.

library/Freckle/App/Memcached/Servers.hs
```diff
@@ -0,0 +1,113 @@
...
+setHost auth ss = case uriRegName auth of
+  "" -> ss
+  rn -> ss { Memcache.ssHost = rn }
+
+setPort :: URIAuth -> Memcache.ServerSpec -> Memcache.ServerSpec
+setPort auth ss = fromMaybe ss $ do
+  -- NB. in older versions of memcache, this field is a PortNumber, so we use
+  -- Read to get one from the URIAuth's string port value. In newer versions,
+  -- it's a ServiceName, which is just type String, so this readMay will
+  -- actually return the String as-is. This approach handles both cases without
```
pbrisbin, line 99: I don't think this works actually, `read @String` requires a quoted value. I'll add a test and figure something out; hopefully not CPP :grimacing:

tests/Freckle/App/Memcached/ServersSpec.hs
```diff
@@ -0,0 +1,92 @@
...
+
+spec :: Spec
+spec = do
+  describe "readMemcachedServers" $ do
+    it "requires the correct prefix" $ example $ do
+      void (readMemcachedServers "http://") `shouldSatisfy` isLeft
+      void (readMemcachedServers "memcached://") `shouldSatisfy` isRight
+
+    it "treats empty cavlues as default" $ example $ do
+      readServerSpecs "" `shouldBe` Just [Memcache.def]
```
pbrisbin, line 23: Now that I've introduced `MemcacheClientDisabled` as a value, should this `MEMCACHED_SERVERS=` case result in that (vs `MEMCACHED_SERVERS=memcached://` to get "defaults")? It would make things slightly nigher over in deployment to unconditionally add the variable, but conditionally make it empty for the environments without caching.

tests/Freckle/App/Memcached/ServersSpec.hs
```diff
@@ -0,0 +1,92 @@
...
+defaultHost :: String
+defaultHost = Memcache.ssHost Memcache.def
+
+#if MIN_VERSION_memcache(0,3,0)
+portServiceName :: String -> String
+portServiceName = id
+#else
+portServiceName :: Show a => a -> String
+portServiceName = show
+#endif
```
eborden, line 92: Gross :laughing:

eborden (approved):
This all looks familiar. :+1:

PR #55, opened 2022-05-24
Migrate Bugsnag handling

eborden:
This commit includes a variety of Bugsnag supports for freckle apps.

- Parsing env variables.
- Refined grouping of common exceptions (http, sql).
- Standard `beforeNotify` hooks to add refinements.

These have been extracted from Freckle's applications with domain
specific logic removed.

### DIFF
--- a/library/Freckle/App/Bugsnag.hs
+++ b/library/Freckle/App/Bugsnag.hs
@@ -28,6 +28,7 @@ import Control.Monad.Reader (runReaderT)
 import Data.Bugsnag
 import Data.Bugsnag.Settings
 import qualified Data.ByteString.Char8 as BS8
+import Data.List (isInfixOf)
 import Database.PostgreSQL.Simple (SqlError(..))
 import Database.PostgreSQL.Simple.Errors
 import qualified Freckle.App.Env as Env
@@ -141,8 +142,7 @@ asHttpException (InvalidUrlException url msg) = updateExceptions $ \ex -> ex
 -- Marking them as not in-project does this, with little downside.
 --
 maskErrorHelpers :: BeforeNotify
-maskErrorHelpers = setStackFramesInProjectByFile (`notElem` errorHelpers)
-  where errorHelpers = ["library/FrontRow/Core/Exceptions.hs"]
+maskErrorHelpers = setStackFramesInProjectByFile (`isInfixOf` "Exceptions")
 
 -- brittany-disable-next-binding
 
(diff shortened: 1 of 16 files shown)


stack.yaml
```diff
@@ -4,6 +4,8 @@ extra-deps:
   - hspec-junit-formatter-1.1.0.2
   - memcache-0.3.0.1
   - scientist-0.0.0.0
+  - bugsnag-1.0.0.1
+  - bugsnag-wai-1.0.0.1
```
eborden, line 8: Wild guess that I'll need to update the other `stack.yaml` files :smile:

pbrisbin: All but nightly ;) I just got this into Stackage recently so it should show up in the LTS soon too.

library/Freckle/App/Bugsnag.hs
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+-- | Set StackFrame's InProject to @'False'@ for Error Helper modules
+--
+-- We want exceptions grouped by the the first stack-frame that is /not/ them.
+-- Marking them as not in-project does this, with little downside.
+--
+maskErrorHelpers :: BeforeNotify
+maskErrorHelpers = setStackFramesInProjectByFile (`notElem` errorHelpers)
+  where errorHelpers = ["library/FrontRow/Core/Exceptions.hs"]
```
pbrisbin, line 145: This is annoying, and I'm open to the practicality argument, but `freckle-app` really shouldn't know about this list of error helpers from our actual domain-specific app.

pbrisbin (approved):
Excellent. I've been meaning to do this!

library/Freckle/App/Bugsnag.hs
eborden: Yeah, totally missed that little line.

eborden: One option is to have it check for `Exceptions` modules in general.

pbrisbin:
Oh, and don't forget to update all the lock files...

```console
for x in stack-lts-*.yaml; do stack --stack-yaml $x build --dry-run; done
```

Yeah, this stuff's supes annoying.

mjgpy3 (approved):
LGTM -- nice tests!

library/Freckle/App/Bugsnag.hs
eborden: For now I'm just going to codify this convention. If we have a need then we can make it more configurable.

PR #59, opened 2022-06-06
Refactor runCapturedLoggingT

pbrisbin:
- Make the interface `[LogLine]`, like [`runWriterLoggingT`][rwlt]
- Store messages in a `DList`, like `runWriterLoggingT` does
- Update comment to indicate why we can't just use `runWriterLoggingT`

[rwlt]: https://hackage.haskell.org/package/monad-logger-0.3.36/docs/Control-Monad-Logger.html#v:runWriterLoggingT

### DIFF
--- a/freckle-app.cabal
+++ b/freckle-app.cabal
@@ -106,6 +106,7 @@ library
     , containers
     , data-default
     , datadog
+    , dlist
     , doctest
     , ekg-core
     , errors
@@ -294,7 +295,6 @@ test-suite spec
     , lens
     , lens-aeson
     , memcache
-    , monad-logger
     , mtl
     , postgresql-simple
     , unliftio-core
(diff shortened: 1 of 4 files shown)


library/Freckle/App/Test/Logging.hs
```diff
@@ -1,32 +1,40 @@
...
 import Control.Monad.Logger
+import Data.DList (DList)
+import qualified Data.DList as DList
 import UnliftIO.Async
 import UnliftIO.Exception (finally)
 
 -- | Run a 'LoggingT', capturing and returning any logged messages alongside
 --
--- I do not know why 'runChanLoggingT' exists presumably for this purpose, but
--- requires so much more effort to ultimately accomplish.
```
pbrisbin, line 15: Apparently, it's because you should be using `runWriterLoggingT`. And screw you if you need `MonadUnliftIO`. :joy:

freckle-app.cabal
```diff
@@ -106,6 +106,7 @@ library
     , containers
     , data-default
     , datadog
+    , dlist
```
mjgpy3, line 109: TIL `DList`, thanks Pat.

PR #60, opened 2022-06-06
Run Database pool creation in MonadLogger

pbrisbin:
This is how it is meant to work: creation of the pool captures the
logging context and will use it to log database activity. Doing this
allowed us to address some TODOs for logging token handling as well. I
also removed some TODOs that were _TODONE_ already, and some dead code.

To recover the previous behavior, end-users can import and use
`runNoLoggingT` themselves; but they are encouraged to use their real
logging context instead.

### DIFF
--- a/freckle-app.cabal
+++ b/freckle-app.cabal
@@ -136,7 +136,6 @@ library
     , persistent-postgresql
     , postgresql-simple
     , primitive
-    , process
     , resource-pool
     , retry >=0.8.1.0
     , rio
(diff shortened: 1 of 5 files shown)


freckle-app.cabal
```diff
@@ -148,6 +147,7 @@ library
     , time
     , transformers
     , transformers-base
+    , typed-process
```
pbrisbin, line 150: I decided to move to `typed-process` as part of this. It allowed fewer `liftIO`s in the DB stuff since it runs in `MonadIO` already. We only had two cases, that both translated well, and the library is excellent in general.

PR #64, opened 2022-06-10
Replace Freckle.App.Logging with Blammo

pbrisbin:
This will require the following end-user changes:

- Remove `HasLogging` instance and add `HasLogger`

  This will require storing a `Logger` on your runtime `App` type,
  instead of a `LogLevel` (et al) in your static `AppSettings` type. (Or
  move to `runSimpleLoggingT`.)

- Replace WAI middleware with integration according to Blammo README

  https://github.com/freckle/blammo/#integration-with-wai

- Replace Yesod functions with integration according to Blammo README

  https://github.com/freckle/blammo/#integration-with-yesod

And changes to environment variable values:

- If using `LOG_OUTPUT=std(out|err)`, use `LOG_DESTINATION=std(out|err)`
- If using `LOG_OUTPUT=file`, use `LOG_DESTINATION=@fancy.log`
- If using `LOG_FORMAT=terminal`, use `LOG_FORMAT=tty`

We drop support for ghc-8.6 and 8.4 here because monad-logger-aeson, and
so Blammo, don't support that far back.

### DIFF
--- a/tests/Freckle/App/MemcachedSpec.hs
+++ b/tests/Freckle/App/MemcachedSpec.hs
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+
 module Freckle.App.MemcachedSpec
   ( spec
   ) where


eborden (approved):
Looks pretty straight forward to me. Nothing is really lost in removing some version of GHC.

pbrisbin:
```
   1) Freckle.App.Memcached.caching logs, but doesn't fail, on deserialization errors
       expected: Just "[Caching] error deserializing: invalid: \"Broken\""
        but got: Just "{\"text\":\"Error deserializing\",\"meta\":{\"action\":\"deserializing\",\"message\":\"invalid: \\\"Broken\\\"\"}}"
  To rerun use: --match "/Freckle.App.Memcached/caching/logs, but doesn't fail, on deserialization errors/"
```

Oh hey, that's cool!

PR #65, opened 2022-06-13
Add Freckle.App.Env.kept

pbrisbin:
Modifies a `Parser` such that all variables act as if they were defined
using `keep`. The past-tense naming was chosen to match the similar
function `prefixed`.

By default, read variables are removed from the environment. This is
often problematic (e.g. in tests that repeatedly load an app and re-read
the environment), and the security benefits are minor. Doing this
variable by variable is error-prone and often outside of the control of
where it matters (e.g. parser defined in this library, ultimately used
in some app). This function will make it easy for the app to decide at
parse time if it wants all variables kept all the time.

### DIFF
--- a/library/Freckle/App/Env.hs
+++ b/library/Freckle/App/Env.hs
@@ -27,6 +27,7 @@ module Freckle.App.Env
   , flag
 
   -- * Extensions
+  , kept
   , eitherReader
   , time
   , keyValues
@@ -39,6 +40,8 @@ import qualified Data.Text as T
 import Data.Time (defaultTimeLocale, parseTimeM)
 import Env hiding (flag)
 import qualified Env
+import Env.Internal.Free (hoistAlt)
+import Env.Internal.Parser (Parser(..), VarF(..))
 
 -- | Designates the value of a parameter when a flag is not provided.
 newtype Off a = Off a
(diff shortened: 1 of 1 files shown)


pbrisbin:
NOTE: `envparse-0.5.0` _just_ released, which makes this unnecessary (they flipped the default behavior), but I think I'd still like to ship this to avoid bringing in the updated `extra-dep` ahead of time.

CHANGELOG.md
```diff
@@ -1,4 +1,21 @@
...
+
+## [v1.2.0.1](https://github.com/freckle/freckle-app/compare/v1.2.0.0...v1.2.0.1)
+
+- Use `Env.kept` with parsing that occurs in `makePostgresPool`
+
+  This ensure all `PG` variables are kept in the environment after parsing, to
+  again match pre-v1.1 behavior.
+
+  If you would prefer not to keep these variables, parse them yourself (e.g.
+  with `envParseDatabaseConf`) and use `makePostgresPoolWith`.
```
mjgpy3, line 11: Would this be more useful in the haddock?

Personally I look there before changelogs for usage info like this.

mjgpy3 (approved):
LGTM, take or leave the doc comment.

CHANGELOG.md
pbrisbin: I'm not sure. To me, this behavior is only notable _because it changed_ (hence CHANGELOG). The behavior as it was/is isn't particularly surprising to me, not worth calling out in docs at least. :shrug:

PR #69, opened 2022-07-06
Re-implement StatsD interface

pbrisbin:
This creates two new abstractions to replace existing ones:

- `Freckle.App.Stats` replaces `Freckle.App.Datadog`

  1. Simplify down to a single capability class, and make it lens-based
     so we can now use `local` to specify tags, rather than force all
     the metric functions to accept a new almost-always unused argument

  2. Expose only a bracket-like usage, to avoid forgotten finalizations

     This automatically adds the tags to the thread context too, which
     ensures that logging and metrics match, for correlation in Datadog.

  3. And fix a bunch of bad naming

- `Network.Wai.Middleware.Stats.requestStats` replaces
  `makeRequestMiddleware`

  To use the new `Stats` internals and fix the incorrect "make"-style
  naming.

Once we migrate over, we can remove the old code in a major version
bump.


library/Freckle/App/Stats.hs
```diff
@@ -0,0 +1,271 @@
...
+
+instance HasStatsClient StatsClient where
+  statsClientL = id
+
+withStatsClient
+  :: (MonadMask m, MonadUnliftIO m)
+  => StatsSettings
+  -> (StatsClient -> m a)
+  -> m a
+withStatsClient StatsSettings {..} f = do
```
pbrisbin, line 160: This usage has a lot of benefits, but unfortunately we expect a `loadApp :: IO App` in a lot of places (e.g. `Test.withApp`), and you can't make such a function if you only have access to a `StatsClient` in a block like this.

I'll have to :thinking: on this.

PR #71, opened 2022-07-08
Start common assertions module

mjgpy3:
It's generally good to show that our `JSON` and `PathPiece` instances can read what they write. We test this sometimes but it's always ad hoc and thus inconsistent (in form and coverage).

The hope is that allowing easy import of these tests will
 1. reduce a little boilerplate
 2. encourage us to write more of these
 3. make this testing pattern consistent, discoverable, and decrease the "reinventing the wheel syndrome"


library/Freckle/App/Test/CommonTests.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Freckle.App.Test.CommonTests
```
mjgpy3, line 4: Another thing I considered adding here was `describeUniverse`....

This would be modeled after our `describeContentArea` function which looks like
```haskell
describeAllContentAreas :: (ContentArea -> SpecWith a) -> SpecM a ()
describeAllContentAreas tests = do
  for_ universe $ \contentArea ->
    describe ("for content area " <> show contentArea) $ tests contentArea
```

mjgpy3: Thoughts?

pbrisbin (approved):

library/Freckle/App/Test/CommonTests.hs
```diff
@@ -0,0 +1,34 @@
...
+-- | Property test that @fromJSON (toJSON value)@ is @value@
+jsonInstancesRoundTrip
+  :: forall a
+   . (Arbitrary a, ToJSON a, FromJSON a, Show a, Eq a)
+  => SpecWith ()
+jsonInstancesRoundTrip =
+  describe "JSON instances"
+    $ it "are inverses (serialization then deserialization round-trips)"
+    $ property
+    $ \(value :: a) -> fromJSON (toJSON value) `shouldBe` Success value
```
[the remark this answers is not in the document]
mjgpy3: Thanks for reasoning this out, Pat! I agree that `Common*` felt smelly and I think your conclusion is "chef's kiss" :man_cook: :kiss:.

pbrisbin (approved):

PR #72, opened 2022-07-11
Require a bracketed loadApp function

pbrisbin:
Currently freckle-app, expects a function like:

```hs
loadApp :: IO App
```

Which it uses in `Freckle.App.runApp` and `Freckle.App.Test.withApp`.
This is problematic for Apps that make use of resources that only expose
their own usage in a bracket-like way:

```hs
-- withStatsClient :: (StatsClient -> m a) -> m a

data App = App
  { ...
  , appStatsClient :: StatsClient
  }

loadApp :: IO App
loadApp = do
  withStatsClient $ \appStatsClient -> do
    -- ???
```

And even if we could work around it, we'd be losing any safety for
resource-cleanup that that usage is intending to provide.

To support this, we can expect a bracket-like function ourselves,

```hs
loadApp :: (App -> IO a) -> IO a
loadApp f = do
  withStatsClient $ \appStatsClient -> do
    f App {..}
```

No problem.

So this commit updates `runApp` and `Test.withApp` to do exactly that,
then updates the Memcache test to take advantage of it, mostly as an
example.

I chose to do this in a backwards-incompatible way rather than make
alternative functions. This method is trivial to implement if you're
using the old way (and the reverse is impossible).

```hs
loadApp :: (App -> IO a) -> IO a
loadApp f = do
  app <- loadApp'
  f app

loadApp' :: IO App
loadApp' = do
  -- your old code
```

I had to update the main documentation header, so I also cleaned up some
out of date, but unrelated aspects therein.

### DIFF
--- a/library/Freckle/App.hs
+++ b/library/Freckle/App.hs
@@ -22,28 +22,27 @@
 -- > instance HasLogger App where
 -- >   loggerL = lens appLogger $ \x y -> x { appLogger = y }
 --
--- and a way to build a value:
+-- and a bracketed function for building and using a value:
 --
--- > loadApp :: IO App
+-- > loadApp :: (App -> m a) -> m a
+-- > loadApp f = do
+-- >   app <- -- ...
+-- >   f app
 --
 -- It's likely you'll want to use @"Freckle.App.Env"@ to load your @App@:
 --
 -- > import qualified Blammo.Logger.LogSettings.Env as LoggerEnv
 -- > import qualified Freckle.App.Env as Env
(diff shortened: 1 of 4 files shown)


pbrisbin:
In the future, I'd love to make canonical usage be that end-users define a `withApp` (instead of `loadApp`), because that's way better naming, but omg the level of effort in migrating all your specs using `Freckle.App.Test.withApp` is just too much for me to think about.

mjgpy3 (approved):

PR #73, opened 2022-07-12
Re-implement StatsD interface

pbrisbin:
This creates a new, simpler/better/faster abstraction to replace the
existing `HasDogStats` one.

1. It's a single class for the client and configured tags
2. The tags parser handles `DD_*` variables automatically
3. Expose only the safer "with"-style usage
4. Said usage automatically,

   - Looks for ECS metadata and adds tags
   - Adds all tags to the thread context (shows up on logs)

5. Individual metrics don't accept extra tags argument

   The `tagged` function can be used to add tags on all metrics emitted
   from a block, which is way more powerful. It can just as easily be
   used for a single metric, but also around something much larger in
   scope.

Internal dependents were also migrated over. The request stats
middleware was moved and renamed to fix a long-standing naming wart.


library/Freckle/App/Stats.hs
```diff
@@ -0,0 +1,280 @@
...
+  -> m ()
+histogramSince = histogramSinceBy toSeconds where toSeconds = round @_ @Int
+
+histogramSinceMs
+  :: (MonadUnliftIO m, MonadReader env m, HasStatsClient env)
+  => Text
+  -> UTCTime
+  -> m ()
+histogramSinceMs = histogramSinceBy toMilliseconds
+  where toMilliseconds = (* 1000) . realToFrac @_ @Double
```
mjgpy3, line 162: Feeling a little [daft][daft]... Please walk me through why we multiply here instead of divide.

[daft]: https://www.youtube.com/watch?v=yydNF8tuVmU

library/Freckle/App/Stats.hs
```diff
@@ -0,0 +1,280 @@
...
+--
+--     'runApp' app $ ...
+--   @
+--
+-- - Throughout your application code, emit metrics as desired
+--
+--   @
+--   import qualified Freckle.App.Stats as Stats
+--
+--   myFunction :: (MonadIO m, MonadReader env m, 'HasStatsClient' env) => m ()
```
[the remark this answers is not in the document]
pbrisbin: Since I'm using `@` to delimit the block (instead of prefixing long lines with `>`), this whole thing is a multiline, quotes-within-monospace.

library/Freckle/App/Stats.hs
pbrisbin: We already measure the duration in seconds. So if it took 5 seconds, we get `5` here and we want to report that as 5,000 ms, `5 * 1000`.

pbrisbin: This should probably be updated to use the monotonic stuff Evan used in Scientist, btw.

library/Freckle/App/Stats.hs
```diff
@@ -0,0 +1,280 @@
+{-# LANGUAGE TupleSections #-}
+
+-- | An intentionally-leaky StatsD interface to Datadog
+--
+-- $usage
```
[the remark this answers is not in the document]
pbrisbin: It's silly, but I very often navigate to top-of-file to get to the exports or imports. Having a massive top-level comment makes that super annoying, so I just used the named block to stash it end-of-file instead.

library/Freckle/App/Stats.hs
mjgpy3: Oh nice, gotcha!

library/Freckle/App/Stats.hs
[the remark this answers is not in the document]
mjgpy3: Gotcha!

PR #74, opened 2022-07-14
Use older primitive extra-dep

pbrisbin:
When I added this, I mistakenly used the latest primitive (0.7.4.0) and
not the minimum as required by aeson (0.7.0.1). This is particularly
problematic because our oldest snapshot sets our lower bounds on
release, so we ended up shipping with a very high lower-bound on
primitive and we're being held out of Stackage because of it.


pbrisbin:
Fixes commercialhaskell/stackage/#6585.

PR #78, opened 2022-09-06
Import xray and ekg support into runDB

tomcarste:
[Consolidate runDB](https://app.asana.com/0/399653121150251/1195508401477809/f)

In order to use this runDB as the api's we need xray and ekg support, so add that here.


library/Freckle/App/Database.hs
```diff
@@ -57,6 +62,9 @@ class HasSqlPool app where
 instance HasSqlPool SqlPool where
   getSqlPool = id
 
+class HasActiveConnectionGauge app where
+  getActiveConnectionGauge :: app -> Gauge.Gauge
```
eborden, line 66: Is there a case where we are okay with this failing? E.G. not having a gauge, or should we just expect all our apps to monitor their active connections?

library/Freckle/App/Database.hs
```diff
@@ -65,12 +73,68 @@ makePostgresPool = do
...
+-- Together, this leads to the nice composition:
+--
+-- @
+-- mVaultData <- getVaultData
+-- maybe 'runSqlPool' ('runSqlPoolXRay' "name") mVaultData action pool
+-- @
+--
+-- N.B. This could make sense in an aws-xray-client-persistent package.
+--
+runSqlPoolXRay
```
eborden, line 116: Similar question on `Xray`. Should we expect all our apps to have xray? Does this fail gracefully? What happens when `Xray` doesn't exist?

tomcarste: Both EKG and Xray were called out as desired everywhere in the task. @pbrisbin

library/Freckle/App/Database.hs
tomcarste: For ease of implementation elsewhere I could definitely see this as being a `Maybe Gauge`, those as below I'm uncertain what the architecture requirements are intended to be.

library/Freckle/App/Database.hs
```diff
@@ -20,8 +20,6 @@ module Freckle.App.Database
   , envPostgresPasswordSource
   ) where
 
-import Freckle.App.Prelude
-
```
pbrisbin, line 24: :art: This should stay up here. We separate a module's prelude import like this conventionally.

library/Freckle/App/Database.hs
pbrisbin: I'd go with `Maybe` to make this optional.

The goal is `runDB` should be able to Do All The Things but be easy to use, so we naturally use it consistently. So if you don't need/want the metrics, that should still be an easy use-case.

The same applies to `HasStatsClient`, but (I believe) that module provides a `DummyClient` that can be used to opt-out but still provide the interface.

library/Freckle/App/Database.hs
pbrisbin: It all should be opt-in.

In this case, I believe `vaultDataFromRequest` is the magic sauce. If the end-user [hasn't done what's required](https://github.com/freckle/megarepo/blob/f701070c8b1ee878fd4afe32d9be0dbec90e41d1/backend/fancy-api/library/Application.hs#L336) for that to work, you hit the `runSqlPool` case. The docs could probably make that clearer, that the WAI Middleware is required before in-band traces (i.e. this `runDB`) will function.

library/Freckle/App/Database.hs
```diff
@@ -57,6 +62,9 @@ class HasSqlPool app where
 instance HasSqlPool SqlPool where
   getSqlPool = id
 
+class HasActiveConnectionGauge app where
```
pbrisbin, line 65: I think you forgot to add this to exports. Gonna be hard to end-users to make the required instance. Personally, I'd also (re)export `Guage` and whatever constructor functions are required to make one. I'm not sure if the qualified import makes that annoying though.

pbrisbin:
I don't know how feasible this is, but I wonder if there's any way to only need `HasStatsClient`. It feels like these two classes are super related, and we only need the separate `Stats.Gauge` one because it's more state that needs defining statically. Would it be terrible if `HasStatsClient` just had some "stock" gauges that got used in other parts of `freckle-app`, with `database.active_connections` being one of them?

library/Freckle/App/Database.hs
tomcarste: We'd need to change restyled for that, is there an annotation I can use?

pbrisbin: Can you show be Restyled doing this?

If you combine the groups, it'll sort them (stylish-haskell), but (as far as I know) it will not combine the groups on you.

tomcarste:
@pbrisbin 
The biggest impediment to HasStatsClient providing say a `withGauge` function would be a circular dependency between Stats and Gauge. Also as far as I can tell we don't directly provide a dummy for StatsClient but Datadog does so it's trivial to wrap.

Having said that, since these are type classes we own it's not difficult to make HasStatsClient be complete enough to be the sole dependency to provide EKG. The big question is just what does an ideal interface look like? Maybe a type class with a type family for metric types, a function for getting those metrics, and a sendMetrics function that wraps some action and sends all the metric types.

library/Freckle/App/Database.hs
tomcarste: You can see it in the PR restyled created. Though as you point out, this is likely because the blank line in between groups ended up with something on it.

pbrisbin: Yup. This was the commit Restyled (stylish-haskell) saw,



So you made it one big group, then it got sorted. If you put it back to separate, we'll be good.

pbrisbin:
> circular dependency between Stats and Gauge

Ah, I didn't consider that.

> Also as far as I can tell we don't directly provide a dummy for StatsClient but Datadog does so it's trivial to wrap

Yeah, that's what we do; it's triggered by the lack of the enabled flag,

```hs
withStatsClient StatsSettings {..} f = do
  if amsEnabled
    then do
      tags <- (amsTags <>) <$> getEcsMetadataTags

      Datadog.withDogStatsD amsSettings $ \client ->
        -- Add the tags to the thread context so they're present in all logs
        withThreadContext (map toPair tags)
          $ f StatsClient { scClient = client, scTags = tags }
    else f $ StatsClient { scClient = Datadog.Dummy, scTags = amsTags } -- <- Right here
```

> The big question is just what does an ideal interface look like? Maybe a type class with a type family for metric types, a function for getting those metrics, and a sendMetrics function that wraps some action and sends all the metric types.

Hmm, while I don't have a concrete picture of this suggestion, I think "type family" is a bit too fancy for me. I think the ideal interface would be,

```diff
 data StatsClient = StatsClient
   { scClient :: Datadog.StatsClient
   , scTags :: [(Text, Text)]
+  , scActiveDatabaseConnectionsGauge :: Gauge
+  , scSomeOtherGauge :: Gauge
   }
```

Or, if you prefer more structure:

```diff
 data StatsClient = StatsClient
   { scClient :: Datadog.StatsClient
   , scTags :: [(Text, Text)]
+  , scGauges :: StatsClientGauges
   }

+ data StatsClientGauges = StatsClientGauges
+   { scgActiveDatabaseConnections :: Gauge
+   , scgSometOther :: Gauge
+   }
```

But, as you say, the circular dependency is the issue, so I might tackle that directly. We could move `HasStatsClient` and `StatsClient` to `Stats.Class.hs`, which can sit below `Stats.hs` and `Stats.Gauge.hs` ?

pbrisbin:
So, here's a wild interface idea... What if `StatsClient` held onto an `IORef (HashMap Text Gauge)` that was initially empty. There'd only be one way to use it,

```hs
withGauge "db.active-connections" $ do
  -- ...
```

This would, initialize `{ "db.active-connections" => newGauge }` if necessary, and do the increment-block-decrement dance. This would pretty much obviate 90% of the `Stats.Gauge` module anyway, so we could just in-line this into `Stats.hs` to prevent any cycles. It'd be a new-reduction in code, by a lot, I think. If/when the Client is holding Gauges, the `Gauge.tag` element is redundant anyway, so we'd be replacing a hard-coded `(Text, EKG.Gauge)` with a dynamically created `HashMap` of the same. From and end-user perspective, the interface is really simple and flexible, IMO.

I guess the question would be, do we ever make use of `Stats.Gauge`s other functionality, or was that always YAGNI? Also, not sure how you/we fail about the lazy-loaded `IORef` pattern.

tomcarste:
Is there any need to have an `IORef`? The `HashMap` could just be initialized empty in a smart constructor.

Also given the ticket that exists would it make sense to push forward with it or should we spend time figuring out this more ideal interface? Just thinking that this direction is a larger change than the ticket originally called for.

pbrisbin:
> Is there any need to have an `IORef`?

I think it has to be mutable, otherwise how do you insert the new gauges the first time they're needed?

> Just thinking that this direction is a larger change than the ticket originally called for.

That's a fair point for "real" work, but this is a debt task right? I wouldn't feel too good leaving more debt in the process. Assuming we agree on the desired direction, that is.

tomcarste:
Ah, yes. It's attached to the App so it has to be mutable at the attachment point since we can't just replace the App. Makes sense. I'm sure there's a better Ref type, but `IORef` seems fine by me.

Seems like the only open question with that interface is if we're using any of Gauge's other functionality.

pbrisbin:
The only usage I can find is what you're doing in this PR,

```hs
      (bracket_ (increment app) (decrement app) action)
      pool
   where
    increment = withGauge Gauge.increment
    decrement = withGauge Gauge.decrement
    withGauge f = runReaderT $ f =<< asks appActiveConnectionGauge
```

I don't think there's any reason we _must_ expose the separate `increment`/`decrement` for the caller to build their own `bracket_`, but I could be wrong.

tomcarste:
It looks like Gauge.hs itself only ever passes through the tags and name fields. Name would presumably be our Map key, so that just leaves the question of if Gauges ever have tags that differ from the StatsClient. I don't know enough about what the tags are for as there's no explanation to answer that.

pbrisbin:
Yeah, I can confirm there is no need for `Gauge.tags`. I'd expect `HashMap Text EKG.Gauge` to be all we need.

> If/when the Client is holding Gauges, the Gauge.tag element is redundant anyway, so we'd be replacing what's then a small static `(Text, EKG.Gauge)` with a dynamically-created `HashMap` of the same.

tomcarste:
Then yeah, I think we can just elide Gauge.hs and embed that HashMap in StatsClient.

tomcarste:
@pbrisbin one implementation prototype served.

library/Freckle/App/Database.hs
```diff
@@ -65,12 +71,68 @@ makePostgresPool = do
...
+    runSqlPool
+    (runSqlPoolXRay "runDB")
+    mVaultData
+    (bracket_ (increment app) (decrement app) action)
+    pool
+  where
+    increment = runReaderT $ Stats.withGauge "db.active-connections" (withEKG (`EKG.add` 1))
+    decrement = runReaderT $ Stats.withGauge "db.active-connections" (withEKG (`EKG.subtract` 1))
+    withEKG f gauge = do
+      liftIO $ f gauge >> EKG.read gauge
```
pbrisbin, line 96: I think it's important not to leak the `EKG` stuff. It's also a bug risk to have to type this name twice. Can something like this work?

```hs
withGauged "db.active-connections" $ do
  maybe
    runSqlPool
    (runSqlPoolXRay "runDB")
    mVaultData
    action
    pool
```

Where `withGauged` is a single encapsulating function in `Freckle.App.Stats`.

I figured I'd raise this in prototype stage, because I _assume_ you'd have to move the wrapper to outside of runner there. In other words, this won't work:

```hs
  maybe
    runSqlPool
    (runSqlPoolXRay "runDB")
    mVaultData
    (withGauged "db.active-connections" action)
    pool
```

Becase `withGauged` needs `MonadReader env m, HasStatsClient env` and `SqlPersistT m` can't supply that. And the bracket-like nature of it means a `lift` can't help either.

There may be a way to get around that, which would be ideal, because I don't know if there are any negative effects to wrapping the increment/decrement outside of, instead of within, the runner.

WDYT?

tomcarste: If we only intend to use withGauged with increment/decrement then I think we could wrap the entire maybe.

So to be clear, you think EKG should be internal to Stats and not leak out? That's probably a good goal.

pbrisbin: > So to be clear, you think EKG should be internal to Stats and not leak out? 

Yup. It's not any kind of deal breaker, but it seems attainable to keep that an implementation detail so I think we should.

tomcarste:
@pbrisbin alright anything else?

library/Freckle/App/Database.hs
```diff
@@ -65,12 +70,64 @@ makePostgresPool = do
   makePostgresPoolWith conf
 
 runDB
-  :: (HasSqlPool app, MonadUnliftIO m, MonadReader app m)
+  :: ( HasSqlPool app
+     , Stats.HasStatsClient app
```
pbrisbin, line 74: We tend to do a "type un-qualified, functions qualified" thing, just like you would with `Text` or `HashMap`,

```hs
import Freckle.App.Stats (HasStatsClient)
import qualified Freckle.App.Stats as Stats
```

library/Freckle/App/Database.hs
```diff
@@ -65,12 +70,64 @@ makePostgresPool = do
...
-  runSqlPool action pool
+  mVaultData <- vaultDataFromRequest <$> waiRequest
+  Stats.withGauge "db.active-connections" $
+    maybe
+      runSqlPool
+      (runSqlPoolXRay "runDB")
+      mVaultData
+      action
+      pool
+
```
pbrisbin, line 91: :art: 
```suggestion
```

library/Freckle/App/Stats.hs
```diff
@@ -15,7 +15,7 @@ module Freckle.App.Stats
   , tagsL
   , withStatsClient
   , HasStatsClient(..)
-
+  , withGauge
```
pbrisbin, line 18: I might put this at the end of the _Reporting_ section.

library/Freckle/App/Stats.hs
```diff
@@ -111,14 +115,33 @@ withStatsClient StatsSettings {..} f = do
...
-
+      gauges <- liftIO $ newIORef mempty
       Datadog.withDogStatsD amsSettings $ \client ->
         -- Add the tags to the thread context so they're present in all logs
         withThreadContext (map toPair tags)
-          $ f StatsClient { scClient = client, scTags = tags }
-    else f $ StatsClient { scClient = Datadog.Dummy, scTags = amsTags }
+          $ f StatsClient { scClient = client, scTags = tags, scGauges = gauges }
+    else do
+      gauges <- liftIO $ newIORef mempty
```
pbrisbin, line 124: If you're doing this in both branches, you could lift it up before the `if amsEnabled`.

library/Freckle/App/Stats.hs
```diff
@@ -111,14 +115,33 @@ withStatsClient StatsSettings {..} f = do
...
+  where
+    createNewEKG ref hashmap = do
+      ekg <- EKG.new
+      liftIO $ writeIORef ref (HashMap.insert name ekg hashmap)
+      pure ekg
+    bracketEKG action value = do
+      liftIO $ value `EKG.add` 1
+      result <- action
+      liftIO $ value `EKG.subtract` 1
+      pure result
```
pbrisbin, line 143: You're calling this "bracket" but you're not actually bracketing it; in other words, an exception in `action` won't decrement the gauge. I think you want something like,

```suggestion
    bracketEKG action value = do
      liftIO $ value `EKG.add` 1
      action `finally` liftIO (EKG.subtract 1)
```

library/Freckle/App/Stats.hs
```diff
@@ -111,14 +115,33 @@ withStatsClient StatsSettings {..} f = do
...
+withGauge name f = do
+  gaugesRef <- asks $ scGauges . view statsClientL
+  gaugesMap <- liftIO $ readIORef gaugesRef
+  gaugeValue <- liftIO $ maybe (createNewEKG gaugesRef gaugesMap) pure (HashMap.lookup name gaugesMap)
+  bracketEKG f gaugeValue
+  where
+    createNewEKG ref hashmap = do
+      ekg <- EKG.new
+      liftIO $ writeIORef ref (HashMap.insert name ekg hashmap)
+      pure ekg
```
pbrisbin, line 138: There are a few ways to get rid of all the repeated `liftIO`s. The one I think I would use is to use `UnliftIO.IORef` which is already lifted.

```suggestion
  gaugesRef <- asks $ scGauges . view statsClientL
  gaugesMap <- readIORef gaugesRef
  gaugeValue <- maybe (createNewEKG gaugesRef gaugesMap) pure (HashMap.lookup name gaugesMap)
  bracketEKG f gaugeValue
  where
    createNewEKG ref hashmap = do
      ekg <- liftIO EKG.new
      ekg <$ writeIORef ref (HashMap.insert name ekg hashmap)
```

library/Freckle/App/Stats.hs
```diff
@@ -111,14 +115,33 @@ withStatsClient StatsSettings {..} f = do
...
+    else do
+      gauges <- liftIO $ newIORef mempty
+      f $ StatsClient { scClient = Datadog.Dummy, scTags = amsTags, scGauges = gauges }
   where toPair = bimap (fromString . unpack) String
 
+withGauge :: (MonadReader app m, HasStatsClient app, MonadUnliftIO m) => Text -> m a -> m a
+withGauge name f = do
+  gaugesRef <- asks $ scGauges . view statsClientL
+  gaugesMap <- liftIO $ readIORef gaugesRef
+  gaugeValue <- liftIO $ maybe (createNewEKG gaugesRef gaugesMap) pure (HashMap.lookup name gaugesMap)
```
pbrisbin, line 132: Uh-oh. I may have led you astray.

This is super duper racy. If multiple threads interleave `readIORef` and `writeIORef`, we would under-increment. Given how hot `runDB` is, that seems like a realistic problem.

Unfortunately, I'm not sure how to solve it. If we were OK with an unnecessary `EKG.new`, you could put the lookup-or-insert inside an `atomicModifyIORef'` to address the race condition, but now you'd be forcibly serializing all `runDB` calls to lock on the Gauge. Again, given how hot `runDB` is, I don't know that that is a good idea to even "ship it and see".

So, I just don't think we can dynamically create gauges. I still think we can keep the single `HasStatsClient` interface with an `scGauges` member. But it would have to be a statically defined structure, either a non-`IORef` `HashMap` or we go ahead with the record type idea presented elsewhere.

I'd be OK with this, since we only use it internally in `freckle-app` now for `runDB`. Giving up the feature of end-users dynamically wrapping things in gauges seems acceptable since we don't even have a use-case for that ourselves.

WDYT?

pbrisbin (commented):
I started leaving some style comments with intention to Approve, but then I hit a problem...

library/Freckle/App/Stats.hs
tomcarste: Ah, I think I meant to actually use bracket and just spaced on it.

library/Freckle/App/Stats.hs
tomcarste: All the racing would happen at start up, but even then I think we'd get issues. I think just a static `scGauge :: Gauge` would be the simplest solution, especially given the lack of use case for more types of gauges.

tomcarste: If you wanted a more complicated interface that allowed apps to hold their gauges abstractly, we could convert it to `scGauge: a` and require `class NamedGauge a where name :: Text` on our `class NamedGauge a => HasStatsClient a`.

pbrisbin: > All the racing would happen at start up

I suppose that's better, but I still think pretty problematic. We boot up a process as part of scaling, it's grabs some N requests at the same time in a thread each, then a ton of them call `runDB` basically immediately. Will that set the gauge up as wrong forever? Will we see negative counts?

> I think just a static `scGauge :: Gauge` would be the simplest solution

Sounds good. Minor nit, but I'd still do `scGauges` (plural), even if the db-connection gauge is the only one we have now.

Adding more (static) gauges later feels likely, _dynamically creating_ gauges is the feature I'm OK to give up.

tomcarste: Yes, it probably would mess up the count forever, you could quite easily lose a bunch of adds and end up negative.

Then you probably want `scGauges :: Gauges` where `data Gauges = { dbConnections :: EKG.Gauge }` so any refactoring down the line doesn't require extensive changes.

pbrisbin: Yeah, that works. I think I called it `StatsClientGauges { scgDbConnections }` originally, just to go with the general theme of the module, but un-prefixed may even be better considering things are generally qualified anyway,

```hs
Stats.withGauge Stats.dbConnections $ do
  ...
```

library/Freckle/App/Stats.hs
tomcarste: I put it next to HasStatsClient because they're used together. If there's a strong preference then I'll move it.

pbrisbin: It's not a strong preference, no, but I am curious about the reasoning: aren't all the reporting functions "used together" with `HasStatsClient`?

library/Freckle/App/Database.hs
```diff
@@ -65,12 +70,63 @@ makePostgresPool = do
...
+--
+-- Together, this leads to the nice composition:
+--
+-- @
+-- mVaultData <- getVaultData
+-- maybe 'runSqlPool' ('runSqlPoolXRay' "name") mVaultData action pool
+-- @
+--
+-- N.B. This could make sense in an aws-xray-client-persistent package.
+--
```
pbrisbin, line 107: This note seems kind crufty. No reason to bring it over into the published package, IMO.

```suggestion
```

library/Freckle/App/Stats.hs
```diff
@@ -108,17 +126,26 @@ withStatsClient
...
         -- Add the tags to the thread context so they're present in all logs
         withThreadContext (map toPair tags)
-          $ f StatsClient { scClient = client, scTags = tags }
-    else f $ StatsClient { scClient = Datadog.Dummy, scTags = amsTags }
+          $ f StatsClient { scClient = client, scTags = tags, scGauges = gauges }
+    else do
+      f $ StatsClient { scClient = Datadog.Dummy, scTags = amsTags, scGauges = gauges }
   where toPair = bimap (fromString . unpack) String
 
+withGauge :: (MonadReader app m, HasStatsClient app, MonadUnliftIO m) => Lens' Gauges EKG.Gauge -> m a -> m a
```
pbrisbin, line 141: This still leaks `EKG.Gauge` in a user-facing type signature. We're so close to _not_ doing that, I think it's worth one last push.

What do you think of:

```hs
module Freckle.App.Stats
  ( ...

  -- * Stateful Gauges
  , GaugeName
  , dbConnections
  , withGauge

  -- * Reporting
  , ...
  ) where

newtype GaugeName = GaugeName
  { unGaugeName :: Gauges -> EKG.Gauge
  }

dbConnections :: GuageName
dbConnections = GaugeName gDBConnections -- rename the actual record accessor

withGauge :: ... => GaugeName -> m a -> m a
withGauge guageName = do
  gauge' <- view $ statsClientL . guagesL . to (unGaugeName guageName)
```

Usage is identical, internals aren't any more complex, and the interface is better (IMO) because we don't leak `EKG.` or inflict our users with `Lens`.

Thoughts?

library/Freckle/App/Stats.hs
```diff
@@ -108,17 +126,26 @@ withStatsClient
...
+withGauge lens' f = do
+  gauge' <- asks $ view (statsClientL . gaugesL . lens')
```
pbrisbin, line 143: :art: 

```suggestion
  gauge' <- asks $ view $ statsClientL . gaugesL . lens'
```

pbrisbin (changes requested):

pbrisbin (approved):

library/Freckle/App/Stats.hs
```diff
@@ -108,17 +133,26 @@ withStatsClient
...
-          $ f StatsClient { scClient = client, scTags = tags }
-    else f $ StatsClient { scClient = Datadog.Dummy, scTags = amsTags }
+          $ f StatsClient { scClient = client, scTags = tags, scGauges = gauges }
+    else do
+      f $ StatsClient { scClient = Datadog.Dummy, scTags = amsTags, scGauges = gauges }
   where toPair = bimap (fromString . unpack) String
 
+withGauge :: (MonadReader app m, HasStatsClient app, MonadUnliftIO m) => GaugeName -> m a -> m a
+withGauge gaugeName f = do
+  gauge' <- asks $ view $ statsClientL . gaugesL . to (unGaugeName gaugeName)
```
pbrisbin, line 150: FYI, `MonadReader` it itself a getter, so you don't need to `asks` a `view`, you can just `view`. Kind of like `MonadState` being a full lens (get and set). Pretty neat.

```suggestion
  gauge' <- view $ statsClientL . gaugesL . to (unGaugeName gaugeName)
```

library/Freckle/App/Stats.hs
```diff
@@ -85,14 +93,31 @@ envParseStatsSettings =
...
 
+-- should be data but newtype for now because there's only one and only one planned
+newtype Gauges = Gauges
+  { gdbConnections :: EKG.Gauge
+  -- ^ Track open db connections
+  }
+
+-- Opaque wrapper to hide EKG.Gauge
+newtype GaugeName = GaugeName
+  { unGaugeName :: Gauges -> EKG.Gauge }
```
pbrisbin, line 104: :art: 
```suggestion
  { unGaugeName :: Gauges -> EKG.Gauge
  }
```

pbrisbin (approved):
Looks great! Thanks, and sorry for all the back and forth.

tomcarste:
Back and forth is good, better now than after having something merged.

PR #88, opened 2022-11-23
Add incremental CSV parsing to freckle-app

eborden:
This code has been migrated out of Freckle's primary application. It fills some gaps in performance and validation in `cassava`.

- Conduit is utilized for stream processing.
- `Control.Monad.Validate` is used for validation and bulk error reporting. This means all errors in a CSV can be reported instead of failing at the first error.
- `CsvException` is exposed to provide better downstream error handling and reporting. As well this type includes an exception for application level validations and sanity checks.
- The `ValidateHeader` class is exposed to allow fast failure and better reporting on missing header values.


mjgpy3 (approved):
LGTM

eborden:
LOL, looks like I need to fix some extra deps.

image

.github/workflows/ci.yml
```diff
@@ -27,7 +27,6 @@ jobs:
           - stack-lts-18.28.yaml # ghc-8.10
           - stack-lts-16.31.yaml # ghc-8.8
           - stack-lts-14.27.yaml # ghc-8.6
-          - stack-lts-12.26.yaml # ghc-8.4
```
eborden, line 30: I don't think this is controversial.

pbrisbin: :+1: `freckle-app` in particular shouldn't come with high compatibility requirements. It's only when we're talking about a library of ours that is a deps-of-deps situation that dropping support can be a cascading annoyance.

library/Freckle/App/Csv.hs
```diff
@@ -0,0 +1,269 @@
...
+-- | Treat CSV header line as 1
+--
+-- CSVs can break rows over lines, but we don't currently handle that.
+--
+headerLineNumber :: Int
+headerLineNumber = 1
+
+-- | Start counting CSV line numbers from 2
+--
+-- We require headers for all customer CSVs, so, for line-oriented
```
pbrisbin, line 72: "customer"

A little business logic is leaking into our domain-agnostic toolkit here. Not blocking in the slightest, but good to keep an eye one.

pbrisbin (approved):

eborden:
A bunch of fun stuff in the errors here. Some `Monad.Fail` redundancy, some differences between version of `cassava` in error reporting. Looks like I'll be building more locally.

pbrisbin:
I've got a lot of tricks for dealing with this stuff, so if you want any help let me know.

eborden:
:point_up: rebased into nicer commits.

pbrisbin:
Don't forget to remove the `lts-12.26.yaml` and disable it as a required status check. Otherwise, looks green to me.

pbrisbin:
(It's important to clean up that yaml file, otherwise it'll get picked up as the minimum resolver to set lower bounds on release.)

PR #90, opened 2022-11-29
Actually publish Gauge values

pbrisbin:
As written, the `withGauge` machinery was incr/decrementing a stateful
gauge but then just never sending it anywhere.

This commit adds the missing `publish` step to the calls, which
necessitated changing the `Gauges` type slightly to also store the
metric name. I also moved to the seemingly more appropriate `inc` and
`dec` `EKG` functions, vs `add 1` and `subtract 1`. It probably doesn't
matter, but this may handle edge cases better, such as decrementing a
0/negative value.

### DIFF
--- a/library/Freckle/App/Stats.hs
+++ b/library/Freckle/App/Stats.hs
@@ -18,7 +18,7 @@ module Freckle.App.Stats
 
   -- * Gauges
   , Gauges
-  , GaugeName
+  , Gauge
   , dbConnections
   , withGauge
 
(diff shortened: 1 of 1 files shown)


pbrisbin:
I think the signature change in `withGauge` makes this a major bump.

tomcarste (approved):
Explains why gauge names were a thing. Though I still don't remember seeing a publish anywhere, so good catch.

PR #91, opened 2022-11-30
Cut 1.8 for the changes to the withGauge function

tomcarste:
Major version bump for the breaking changes in `withGauge`.

### DIFF
--- a/library/Freckle/App/Database.hs
+++ b/library/Freckle/App/Database.hs
@@ -55,7 +55,8 @@ import System.Process.Typed (proc, readProcessStdout_)
 import UnliftIO.Concurrent (threadDelay)
 import UnliftIO.Exception (displayException)
 import UnliftIO.IORef
-import Yesod.Core (MonadUnliftIO (withRunInIO))
+import Yesod.Core (MonadUnliftIO(withRunInIO), YesodRequest(reqWaiRequest))
+import Yesod.Core.Types (HandlerData(handlerRequest))
 
 type SqlPool = Pool SqlBackend
 
@@ -67,6 +68,9 @@ instance HasSqlPool SqlPool where
 
 class HasVaultData env where
   getVaultData :: env -> Maybe XRayVaultData
+
+instance HasVaultData (HandlerData child site) where
+  getVaultData = vaultDataFromRequest . reqWaiRequest . handlerRequest
 
 makePostgresPool :: (MonadUnliftIO m, MonadLoggerIO m) => m SqlPool
 makePostgresPool = do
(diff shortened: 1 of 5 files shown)


tomcarste:
Decided to go for a stop gap type class similar to HasSqlPool. Mostly because getting the Maybe right in a lens produced an abhorrent instance for XRayVaultData* and not having the Maybe requires changes to the underlying library. In the end our goal is to have our middleware take care of handling non-present VaultData and we can make this a regular lens has field type class.

```vaultDataL f x = fmap (const x) $ f (Just x)```

I'm sure there has to be something better than this.

**Comment by tomcarste** [comment]
Also @pbrisbin any instances that we should supply by default?

File: library/Freckle/App/Database.hs [hunk 1]
```diff
@@ -54,7 +55,7 @@ import System.Process.Typed (proc, readProcessStdout_)
 import UnliftIO.Concurrent (threadDelay)
 import UnliftIO.Exception (displayException)
 import UnliftIO.IORef
-import Yesod.Core (MonadHandler, MonadUnliftIO(withRunInIO), waiRequest)
+import Yesod.Core (MonadUnliftIO (withRunInIO))
```
pbrisbin, line 58: :art: 
```suggestion
import Yesod.Core (MonadUnliftIO(withRunInIO))
```

library/Freckle/App/Test.hs
```diff
@@ -103,6 +103,8 @@ instance HasLogger app => Example (AppExample app a) where
     params
     ($ ())
 
+instance HasVaultData app => HasVaultData (AppExample app a) where
+  getVaultData = asks getVaultData
```
pbrisbin, line 107: :art: 
```suggestion
  getVaultData = asks getVaultData
  
```

pbrisbin (commented):
I think, to use this from a `HandlerFor app`, you'll want:

```hs
instance HasVaultData (HandlerData child site)
  getVaultData = vaultDataFromRequest . reqWaiRequest . handlerRequest
```

library/Freckle/App/Database.hs
```diff
@@ -74,15 +82,15 @@ makePostgresPool = do
...
+     , HasVaultData app
      , MonadUnliftIO m
      , MonadReader app m
      )
   => SqlPersistT m a
   -> m a
 runDB action = do
   pool <- asks getSqlPool
-  mVaultData <- vaultDataFromRequest <$> waiRequest
+  mVaultData <- asks getVaultData-- vaultDataFromRequest <$> waiRequest
```
pbrisbin, line 93:
```suggestion
  mVaultData <- asks getVaultData
```

pbrisbin (approved):

PR #95, opened 2023-03-01
pb/prefactors

pbrisbin:
- Add lts-19.33 file
- Update stack-nightly
- Update package.yaml to align with template
- Fix warning in Spec.hs
- Avoid direct record construction for Dotenv.Config
- Remove TODOs
- Style
- Lint
- Move 8.8 GHC option under conditional
- Update default resolver to lts-20.12/ghc-9.2
- Update CI matrix


library/Freckle/App.hs
```diff
@@ -134,24 +141,63 @@
...
+
+runAppT
+  :: (MonadUnliftIO m, HasLogger app)
+  => (forall b . (app -> m b) -> m b)
+  -> AppT app m a
+  -> m a
+runAppT loadApp action = do
+  setLineBuffering
+  loadApp $ \app ->
+    runResourceT $ runLoggerLoggingT app $ runReaderT (unAppT action) app
```
pbrisbin, line 197: I think I might prefer leaving `runAppT` as the "just unwind the stack" function, and `runApp` as the "set buffering, load app, etc" function...

```hs
runApp loadApp action = do
  setLineBuffering
  loadApp $ runAppT action

runAppT action app =
  runResourceT $ runLoggerLoggingT app $ runReaderT (unAppT action) app
```

Is that less surprising?

library/Freckle/App.hs
```diff
@@ -135,23 +142,62 @@
...
 import Control.Monad.Reader as X
+import Control.Monad.Trans.Resource (MonadResource, ResourceT, runResourceT)
 import Freckle.App.Database as X
 import System.IO (BufferMode(..), hSetBuffering, stderr, stdout)
 
 runApp
   :: HasLogger app
   => (forall b . (app -> IO b) -> IO b)
-  -> ReaderT app (LoggingT IO) a
+  -> AppT app IO a
```
pbrisbin, line 164: Even though this changes, users of `runApp` shouldn't have to change anything, as long as they were writing their actions against `MonadReader` and other constraints, and calling `runApp` without a type-signature on the action.

library/Freckle/App.hs
```diff
@@ -59,6 +59,13 @@
 -- >     myAppAction
 -- >     myOtherAppAction
 --
+-- == 'AppT'
```
StevenXL, line 62: Is it possible to avoid exposing the stack so that clients are forced to use the constraints?

tests/Spec.hs
```diff
@@ -1 +1 @@
-{-# OPTIONS_GHC -F -pgmF hspec-discover -optF --module-name=Spec #-}
+{-# OPTIONS_GHC -F -pgmF hspec-discover -optF --module-name=Spec -Wno-missing-export-lists #-}
```
StevenXL, line 1: I don't understand the need for the new option, but that's non-blocking.

StevenXL (approved):

tests/Spec.hs
pbrisbin: I updated the default warnings, so `missing-export-lists` is now enabled. Since this module is generated, without an export list, it fails unless we silence the warning here.

library/Freckle/App.hs
pbrisbin: Anything is possible!

I can not export `AppT(..)` or `runAppT`. Then users would be unable to do the things I mentioned. They'd be forced to use constraints and only ever call `runApp`.

I dislike it when a function (`runApp`) mentions a type (`AppT`) that isn't exported though because:

- You can't get that type's Haddocks (minor)
- You can't extract/use the function in a way that requires a type signature (medium)

I could solve these by exporting `AppT`, but not `(..)` or `runAppT`. Then it's just about whether the limitations are a feature or a bug. What if they needed to use this `AppT` with a different `m` than `IO`? We'd be disallowing that valid use in an attempt to forbid invalid use.

But then, what use are we forbidding that's invalid? There can be no _current_ uses, since `AppT` doesn't exist. And if there are uses that are explicitly `ReaderT`, they'll be broken if we export `AppT` or not. So we'd be guarding against future explicit uses of `AppT`, in case we want to change it -- which is what I mentioned in the comment.

But, wait, is that event true? I think we _can_ evolve the stack under `AppT` while also exporting it: `AppT(..)` + `runAppT` won't change if we do such things. And, even if it did, we play really loose with compatibility in freckle-app since it's not a core library and is more of a "leaf" package anyway.

So I think I just want to ditch the comment hinting at this being any kind of "internal" thing.

PR #96, opened 2023-03-01
Add Freckle.App.OpenTelemetry

pbrisbin:
And update `runDB` to use it.

A `runDBXRay` is preserved in the (unlikely) event one of our apps needs
to update but cannot satisfy `MonadTracer`. To use it requires silencing
the deprecation warning, which is intentional.

### DIFF
--- a/library/Freckle/App/Database.hs
+++ b/library/Freckle/App/Database.hs
@@ -13,6 +13,7 @@ module Freckle.App.Database
   , makePostgresPool
   , makePostgresPoolWith
   , runDB
+  , runDBXRay
   , runDBSimple
   , PostgresConnectionConf(..)
   , PostgresPasswordSource(..)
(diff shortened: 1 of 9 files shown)


stack.yaml
```diff
@@ -1,3 +1,17 @@
 resolver: lts-20.12
 extra-deps:
   - monad-validate-1.2.0.1
+
+  # For OTel
+  - github: freckle/hs-opentelemetry-awsxray
```
pbrisbin, line 6: I should probably get a version on Hackage before merging this. Will do.

library/Freckle/App/Database.hs
```diff
@@ -89,11 +125,12 @@ runDB
      )
   => SqlPersistT m a
   -> m a
-runDB action = do
+runDBXRay action = do
   pool <- asks getSqlPool
   mVaultData <- asks getVaultData
   Stats.withGauge Stats.dbConnections
     $ maybe runSqlPool (runSqlPoolXRay "runDB") mVaultData action pool
+{-# DEPRECATED runDBXRay "Migrate to MonadTracer and use runDB" #-}
```
pbrisbin, line 133: Am I introducing and immediately deprecating a function? Why yes, I am.

pbrisbin:
Oh sorry, there's some more context [in Confluence](https://renaissancelearning.atlassian.net/wiki/spaces/ARCH/pages/42753688399/XRay+Tracing+via+OpenTelementry) too.

tomcarste (approved):
Looks good, seems we're no longer going to use xray?

pbrisbin:
> seems we're no longer going to use xray?

In the log-term, yeah, but not yet. We're going to emit OpenTelemetry, and collector will export to X-Ray (for now), which creates a seam for exploring replacements (like Datadog) later. The link I posted in my last comment has more details.

PR #97, opened 2023-03-02
Address some OTel omissions

pbrisbin:
### [Update package options](https://github.com/freckle/freckle-app/pull/97/commits/bdc1cefe1da3011b76910a79f345b20e3368ed07)
[bdc1cef](https://github.com/freckle/freckle-app/pull/97/commits/bdc1cefe1da3011b76910a79f345b20e3368ed07)

### [Add necessary instances to AppT](https://github.com/freckle/freckle-app/pull/97/commits/b1d74b1920c24e06a70ec516490db5584110be0c)
[b1d74b1](https://github.com/freckle/freckle-app/pull/97/commits/b1d74b1920c24e06a70ec516490db5584110be0c)

Discovered when upgrading megarepo/backend.

### [Extend Freckle.App.OpenTelementry](https://github.com/freckle/freckle-app/pull/97/commits/4a259e2916b371c7edf8d7a6539e828b24d4eab1)
[4a259e2](https://github.com/freckle/freckle-app/pull/97/commits/4a259e2916b371c7edf8d7a6539e828b24d4eab1)

- Re-export some more types required to initialize
- Add `withTracerProviderXRayLogger`
- Export individual setters and `withTracerProvider`

### DIFF
--- a/library/Freckle/App/OpenTelemetry.hs
+++ b/library/Freckle/App/OpenTelemetry.hs
@@ -6,6 +6,13 @@ module Freckle.App.OpenTelemetry
   , MonadTracer(..)
   , inSpan
 
+  -- * Querying
+  , getCurrentTraceId
+  , getCurrentTraceIdHex
+
+  -- * Middleware
+  , addTraceIdThreadContext
+
   -- * Setup
   -- ** Built-ins
   , withTracerProviderXRay
@@ -26,9 +33,19 @@ module Freckle.App.OpenTelemetry
 import Freckle.App.Prelude
 
 import Blammo.Logging
-  (LogLevel(..), Logger, Message(..), logOtherNS, runLoggerLoggingT)
+  ( LogLevel(..)
+  , Logger
+  , Message(..)
+  , logOtherNS
+  , runLoggerLoggingT
+  , withThreadContext
+  , (.=)
+  )
 import Control.Lens ((.~), (<>~))
+import Network.Wai (Middleware)
 import OpenTelemetry.AWSXRay
+import qualified OpenTelemetry.Context as Trace
+import qualified OpenTelemetry.Context.ThreadLocal as Trace
 import qualified OpenTelemetry.Logging.Core as Trace
 import OpenTelemetry.Trace
   ( HasTracer(..)
@@ -40,6 +57,9 @@ import OpenTelemetry.Trace
   , makeTracer
   , tracerOptions
   )
+import qualified OpenTelemetry.Trace.Core as Trace
+  (SpanContext(..), getSpanContext)
+import OpenTelemetry.Trace.Id (Base(Base16), TraceId, traceIdBaseEncodedText)
 import OpenTelemetry.Trace.Monad (MonadTracer(..))
 import qualified OpenTelemetry.Trace.Monad as Trace
 import OpenTelemetry.Trace.Setup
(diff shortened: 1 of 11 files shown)


mjgpy3:
Looking now.

library/Freckle/App/OpenTelemetry.hs
```diff
@@ -1,41 +1,139 @@
...
+addTraceIdThreadContext :: Middleware
+addTraceIdThreadContext app request respond = do
+  mTraceId <- getCurrentTraceIdHex
+
+  let
+    wrap = case mTraceId of
+      Nothing -> id
+      Just traceId -> withThreadContext ["trace_id" .= traceId]
+
+  wrap $ app request respond
```
mjgpy3, line 80: `traverse` by another name?

library/Freckle/App/OpenTelemetry.hs
```diff
@@ -1,41 +1,139 @@
...
+-- TODO: Use more fields, i.e. metadata
+--
+-- <https://hackage.haskell.org/package/hs-opentelemetry-api-0.0.3.6/docs/OpenTelemetry-Logging-Core.html#t:Log>
+--
+fromOTelLog :: Trace.Log Text -> (LogLevel, Message)
+fromOTelLog = fromOTelSeverity . Trace.severityNumber &&& (:# []) . Trace.body
+
+fromOTelSeverity :: Maybe Int64 -> LogLevel
+fromOTelSeverity = \case
+  Just n | n > 17 -> LevelError
```
mjgpy3, line 126: I always wonder how these severity numbers get picked.

mjgpy3 (approved):
LGTM

pbrisbin:
I'm going to poke at the wait integration a bit more before I merge. I don't expect major changes, but let me know if you'll want a re-review and I'll make it Draft until it's _really_ ready. I'll take silence as permission to hack and merge :)

library/Freckle/App/OpenTelemetry.hs
pbrisbin: I don't now! If you can rework to be, please do, I'd like to see it.

library/Freckle/App/OpenTelemetry.hs
pbrisbin: God bless open standards,

[The spec](https://opentelemetry.io/docs/reference/specification/logs/data-model/#field-severitynumber) links to [the design discussion](https://docs.google.com/document/d/1ix9_4TQO3o-qyeyNhcOmqAc1MTyr-wnXxxsdWgCMn9c/edit#), which links to a [separate discussion](https://docs.google.com/document/d/1ix9_4TQO3o-qyeyNhcOmqAc1MTyr-wnXxxsdWgCMn9c/edit#) which has a wonderful FAQ for that!

> **Why do we need a range of severity_number values with similar meaning?**
>
> Different logging libraries and logging formats define different sets of possible values for severities. Typically, broadly speaking, the severity values fall in the 6 ranges as defined in this document. However, often there are also more nuanced severity grades in source formats. For example Syslog defines 4 different error severities: (Emergency, Alert, Critical, Error). We need to be able to represent all these severities in the OpenTelemetry data model. The range of severity values makes it possible.
>
> **Why do we have so many values in each range?**
>
> The largest number of values in a range I have found so far is Syslog and Zap. Both have 4 different error levels. There may be formats and libraries that have even more levels so I chose to have 8 levels in each range just to be safe. There does not seem to be downsides to allowing more levels (other than requiring an extra bit to represent).
Why do severity_number values start from 1 and not from 0?
>
> In certain data formats (e.g. ProtoBuf) fields with 0 values are omitted on the wire and there is no way to distinguish omitted and 0 values in receivers. It is efficient to reserve 0 value as an indication of an omitted field (otherwise we would need to record the fact that severity_number field is present or omitted separately).

PR #99, opened 2023-03-08
Remove OpenTelemetry module, keep MonadTracer

pbrisbin:
We're not able to move to OpenTelemetry while still using X-Ray. The
reason is that the options for continuing to support our method of only
sending "interesting" HTTP traces to X-Ray are severely limited. If we
were to lose this filtering our X-Ray bill would likely increase from
$19 dollars/month to over $400, which is unacceptable.

We are planning to migrate from X-Ray to Datadog APM, at which point we
will have more control over retention and not need this filtering.
Meaning anything we built in the meantime would be wasted.

Therefore, we're going to hold off on OpenTelemetry until we're ready to
migrate to Datadog APM at the same time.

This commit removes the actual `OpenTelemetry` module, puts our `runDB`
back to being X-Ray-based, but changes the `HasXRayVaultData` class to
`MonadTracer`. This means client code can be updated to use the new kind
of constraint ahead of time (work I've already prepared), and need fewer
changes when `OpenTelemetry` actually does enter the picture.


pbrisbin:
In case it's not clear, I'm backing out this work so that `main` is release-able, usable, and update-to-able in megarepo, in case any other work comes along that needs to do that.

mjgpy3 (approved):
Looks and sounds good to me.

PR #102, opened 2023-05-02
Report problems fetching ECS container metadata

pbrisbin:
This has stopped working at some point. We intentionally silence errors
to `Nothing` so they don't prevent any kind of app start and consider
this best effort. That said, we would still like to debug these, so this
patch upgrades that to an `Either` (via `MonadError`) and prints them on
`stderr`.


tomcarste (approved):
LGTM

PR #103, opened 2023-05-16
Extend debuggability and fix ECS Metadata

pbrisbin:
The first few commits increased visibility in the process further. The last commit actually fixes the bug:

### [Request container metadata without trailing /](https://github.com/freckle/freckle-app/pull/103/commits/ace1f1d6a1bc2154325c391d9ff035f0f79d3ddf)
[ace1f1d](https://github.com/freckle/freckle-app/pull/103/commits/ace1f1d6a1bc2154325c391d9ff035f0f79d3ddf)

At some point this started being a 404 with the trailing `/`. Removing
it gets things working again.

### DIFF
--- a/library/Freckle/App/Stats.hs
+++ b/library/Freckle/App/Stats.hs
@@ -267,8 +267,8 @@ sendMetric metricType name metricValue = do
 
 getEcsMetadataTags :: MonadIO m => m [(Text, Text)]
 getEcsMetadataTags = do
-  emMetadata <- runExceptT getEcsMetadata
-  either (([] <$) . err) (pure . maybe [] toTags) emMetadata
+  eMetadata <- runExceptT getEcsMetadata
+  either (([] <$) . err) (pure . toTags) eMetadata
  where
   err e = liftIO $ hPutStrLn stderr $ "Error reading ECS Metadata: " <> show e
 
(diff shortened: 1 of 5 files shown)


library/Freckle/App/Stats.hs
```diff
@@ -267,8 +267,8 @@ sendMetric metricType name metricValue = do
 
 getEcsMetadataTags :: MonadIO m => m [(Text, Text)]
 getEcsMetadataTags = do
-  emMetadata <- runExceptT getEcsMetadata
-  either (([] <$) . err) (pure . maybe [] toTags) emMetadata
+  eMetadata <- runExceptT getEcsMetadata
+  either (([] <$) . err) (pure . toTags) eMetadata
  where
   err e = liftIO $ hPutStrLn stderr $ "Error reading ECS Metadata: " <> show e
```
pbrisbin, line 273: Maybe this should be softer. I worry it'll make folks think their local instance is broken since it says "Error". Perhaps, _Not configuring tags from ECS Metadata: <reason>_ is better.

mjgpy3: We can probably wait and see

mjgpy3 (approved):
LGTM

PR #105, opened 2023-06-13
Set schema in connection from `PGSCHEMA`

stackptr:
`progress-api` is setup to use a `progress` schema for its tables, rather than the default `public` schema typically used. This requires the `psql` connection to be set with a the `-csearch_path` option.

This works as expected when using a tunnel to the prod `progress-api` db for `psql` locally:

```
> ENV=progress-api ./scripts/run-command-ssh-tunnel.sh psql options=-csearch_path=progress
Null display is "<NULL>".
psql (12.15 (Ubuntu 12.15-1.pgdg22.04+1), server 12.12)
SSL connection (protocol: TLSv1.2, cipher: AES128-SHA256, bits: 128, compression: off)
Type "help" for help.

progress_prod=> \dt
                   List of relations
  Schema  |           Name           | Type  |  Owner
----------+--------------------------+-------+----------
 progress | math_question_parameters | table | postgres
(1 row)
```

Without `options=-csearch_path=progress` set, no relations are found until performing `SET search_path TO progress;`.

### DIFF
--- a/library/Freckle/App/Database.hs
+++ b/library/Freckle/App/Database.hs
@@ -130,6 +130,7 @@ data PostgresConnectionConf = PostgresConnectionConf
   , pccDatabase :: String
   , pccPoolSize :: Int
   , pccStatementTimeout :: PostgresStatementTimeout
+  , pccSchema :: String
   }
   deriving stock (Show, Eq)
 
@@ -199,6 +200,7 @@ envParseDatabaseConf source = do
   database <- Env.var Env.nonempty "PGDATABASE" mempty
   port <- Env.var Env.auto "PGPORT" mempty
   poolSize <- Env.var Env.auto "PGPOOLSIZE" $ Env.def 10
+  schema <- Env.var Env.auto "PGSCHEMA" mempty
   statementTimeout <-
     Env.var (Env.eitherReader readPostgresStatementTimeout) "PGSTATEMENTTIMEOUT"
       $ Env.def (PostgresStatementTimeoutSeconds 120)
@@ -210,6 +212,7 @@ envParseDatabaseConf source = do
     , pccDatabase = database
     , pccPoolSize = poolSize
     , pccStatementTimeout = statementTimeout
+    , pccSchema = schema
     }
 
 data AuroraIamToken = AuroraIamToken
@@ -309,4 +312,5 @@ postgresConnectionString PostgresConnectionConf {..} password =
     , "user=" <> pccUser
     , "password=" <> password
     , "dbname=" <> pccDatabase
+    , "options=-csearch_path=" <> pccSchema
     ]


library/Freckle/App/Database.hs
```diff
@@ -130,6 +130,7 @@ data PostgresConnectionConf = PostgresConnectionConf
   , pccDatabase :: String
   , pccPoolSize :: Int
   , pccStatementTimeout :: PostgresStatementTimeout
+  , pccSchema :: String
```
pbrisbin, line 133: I don't think we should require it. I think we should ether make this `Maybe String`, or set a `Env.def "public"` in the parser.

library/Freckle/App/Database.hs
```diff
@@ -309,4 +312,5 @@ postgresConnectionString PostgresConnectionConf {..} password =
     , "user=" <> pccUser
     , "password=" <> password
     , "dbname=" <> pccDatabase
+    , "options=-csearch_path=" <> pccSchema
```
pbrisbin, line 315: It feels inconsistent to handle the timeout by executing a command on the connection at startup, but handle the schema by pushing `options` into the connection string. Can we make it so we do them by the same method? I don't mind which way -- I assume we can set a timeout via `options` or we can set the schema via executing a command.

pbrisbin (changes requested):
I'm only blocking on making this optional. The other comment you can take-or-leave.

library/Freckle/App/Database.hs
stackptr: I didn't realize that's how timeouts are handled. I think this should be done this same way as that

stackptr:
Do I need to rev `version` in `package.yaml` prior to merge?

library/Freckle/App/Database.hs
```diff
@@ -199,6 +200,7 @@ envParseDatabaseConf source = do
   database <- Env.var Env.nonempty "PGDATABASE" mempty
   port <- Env.var Env.auto "PGPORT" mempty
   poolSize <- Env.var Env.auto "PGPOOLSIZE" $ Env.def 10
+  schema <- Env.var Env.auto "PGSCHEMA" mempty
```
pbrisbin, line 203: `auto` means `Read a => a`; it's good for numeric types. Here, you're using it with `a ~ Maybe String`. In other words, this parser will accept `PGSCHEMA=Just "Something"'` or `PGSCHEMA=Nothing`, which is not what you want.

When you're parsing a text type, there's no reason to `Read` it, you can just use `str`. In this case though, we don't want the variable to be empty either, so I'd use `nonempty`. Getting it to `Maybe` is done via `optional`.

```suggestion
  schema <- optional (Env.var Env.nonempty "PGSCHEMA" mempty)
```

library/Freckle/App/Database.hs
```diff
@@ -270,20 +273,22 @@ refreshIamToken conf tokenIORef = do
...
-setTimeout :: MonadIO m => PostgresConnectionConf -> Connection -> m ()
-setTimeout PostgresConnectionConf {..} conn = do
+setStartupOptions :: MonadIO m => PostgresConnectionConf -> Connection -> m ()
+setStartupOptions PostgresConnectionConf {..} conn = do
   let timeoutMillis = postgresStatementTimeoutMilliseconds pccStatementTimeout
-  void $ liftIO $ execute
-    conn
-    [sql| SET statement_timeout = ? |]
-    (Only timeoutMillis)
+  void $ liftIO $ do
```
pbrisbin, line 279: I think, since you're ending with `for_`, the return type is now `()` already.
```suggestion
  liftIO $ do
```

pbrisbin (changes requested):

pbrisbin:
When to release is up to you. You can bump the version in the PR or after.

pbrisbin (approved):

PR #106, opened 2023-07-20
Add new module for kafka

z0isch:
https://app.asana.com/0/1201325186562318/1204292061062064/f

We'd like to begin producing events to Kafka in `fancy-api` when a student answers a question in adaptive math. 

This adds a `Freckle.App.Kafka` module with a `produceAsync` function constrained by a `HasKafkaProducer` typeclass. 
`HasKafkaProducer` uses a `Pool` of connections in order to make producing events more performant as we are going to be using this in a very hot path in `fancy-api`.


library/Freckle/App/Kafka.hs
```diff
@@ -0,0 +1,132 @@
...
+instance HasKafkaProducer site => HasKafkaProducer (HandlerData child site) where
+  kafkaProducerL = envL . siteL . kafkaProducerL
+
+kafkaProducer
+  :: ( MonadReader env m
+     , HasKafkaBrokerAddress env
+     , MonadUnliftIO m
+     )
+  => Int
+  -- ^ The number of stripes (distinct sub-pools) to maintain.
```
z0isch, line 66: The docs here were stolen wholesale from the underlying `Pool.createPool`. 

I thought about giving some sensible defaults instead of parametrizing this function, but this flexibility for the client seemed nice.

library/Freckle/App/Kafka.hs
```diff
@@ -0,0 +1,132 @@
...
+  :: ( ToJSON value
+     , ToJSON key
+     , MonadLogger m
+     , MonadReader env m
+     , HasKafkaProducer env
+     , MonadUnliftIO m
+     )
+  => Kafka.TopicName
+  -> NonEmpty value
+  -> (value -> key)
```
pbrisbin, line 106: This will be an interesting design point vs a `HasKafkaPartitionKey` class on `value`. On the one hand, the class would make it easy to decide partitioning in one place, where the event data type is created. Just passing a function on every send is pretty easy.

You could also offer all the options though...

```hs
produce :: HasKafkaPartitionKey value => value -> 
produce = produceKeyedBy kafkaPartitionKey

produceKeyedBy :: (value -> key) -> value ->
produceKeyedBy f v = produceKeyed (f v) v

produceKeyed :: key -> value ->
produceKeyed k v = do
```

library/Freckle/App/Kafka.hs
```diff
@@ -0,0 +1,132 @@
...
+  -> m ()
+produceAsync prTopic values keyF = void $ async $ do
```
pbrisbin, line 108: I don't see any downside to starting out with both versions available, as it's so easy to define...

```hs
produce = do

produceAsync = void . async . produce
```

pbrisbin (commented):
Cool stuff!

I think it's a good start, but left some ideas that could be good to think about before any first release.

library/Freckle/App/Kafka.hs
```diff
@@ -0,0 +1,132 @@
...
+import qualified Data.Pool as Pool
+import qualified Data.Text as T
+import qualified Freckle.App.Env as Env
+import qualified Kafka.Producer as Kafka
+import UnliftIO.Async (async)
+import UnliftIO.Exception (throwString)
+import Yesod.Core.Lens
+import Yesod.Core.Types (HandlerData)
+
+newtype BrokerAddress = BrokerAddress { getBrokerAddress :: Maybe (NonEmpty Kafka.BrokerAddress)}
```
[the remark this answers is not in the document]
z0isch: I fought with this too lol and had `BrokerAddresses` initially but then ripped it out- I'll revert that

library/Freckle/App/Kafka.hs
```diff
@@ -0,0 +1,132 @@
...
+  kafkaBrokerAddressL = envL . siteL . kafkaBrokerAddressL
+
+envKafkaBrokerAddress
+  :: Env.Parser Env.Error BrokerAddress
+envKafkaBrokerAddress = BrokerAddress <$> addressOrNothing
+
+ where
+  addressOrNothing = (parseKafkaBrokerAddress <$> parseKey) <|> pure Nothing
+  parseKey = Env.var Env.nonempty "KAFKA_BROKER_ADDRESS" mempty
+  parseKafkaBrokerAddress = NE.nonEmpty . fmap Kafka.BrokerAddress . T.splitOn ","
```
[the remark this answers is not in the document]
pbrisbin: Note that we do have some reusable parsers [here](https://hackage.haskell.org/package/freckle-app-1.9.0.3/docs/Freckle-App-Env.html#g:2). I keep wanting something that parses a comma-separated variable, but I can't figure out the perfect interface for passing in the "and parse each value with this function" part of it.

library/Freckle/App/Kafka.hs
pbrisbin: One option is to name what you have `produceKeyedBy`, so as not to squat on `produce(Async)` for something class-based later.

library/Freckle/App/Kafka.hs
```diff
@@ -0,0 +1,132 @@
...
+  -- intervals.
+  -> Int
+  -- ^ Maximum number of resources to keep open per stripe.  The
+  -- smallest acceptable value is 1.
+  --
+  -- Requests for resources will block if this limit is reached on a
+  -- single stripe, even if other stripes have idle resources
+  -- available.
+  -> m KafkaProducer
+kafkaProducer numStripes idleTime maxResources =
```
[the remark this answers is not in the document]
z0isch: I struggled with this and had `createKafkaProducerPool` initially. I shied away due to the return type being just `m KafkaProducer`. I wonder if that should be `KafkaProducerPool` too?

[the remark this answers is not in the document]
pbrisbin: Oh that is tricky. The persistent functions return `Pool SqlBackend`.

Personally, I would _try_ with it returning `Pool Kafka.KafkaProducer`. This would require refactoring some things that also appeal to me...

Have `envKafkaBrokerAddresses` return `NonEmpty BrokerAddress`, then the user is responsible to use `optional` (or equivalent) and get themselves a `settingsKafkaBrokerAddresses :: Maybe (NonEmpty BrokerAddress)`.

When users initialize things, they'd be responsible to do something like:

```hs
producer <- maybe
  (pure NullKafkaProducer)
  (fmap . KafkaProducerPool createKafkaProducerPool)
  settingsKafkaBrokerAddresses
```

I might also switch your top-level type to:

```hs
data KafkaProducerPool
  = KafkaProducerPool (Pool KafkaProducer)
  | NullKafkaProducerPool
```

Which means less (no?) `qualified` business.

This drops some boilerplate and indirection out of the module itself, without imposing too much on the client, IMO. As I said, I would just try this direction and see. It could be a dead-end for reasons I'm not seeing from here.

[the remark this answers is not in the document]
z0isch: I was going back and forth about how much to make the client do here, I think the design you have here probably strikes a good balance and seems to work nicely in `fancy-api`

library/Freckle/App/Kafka.hs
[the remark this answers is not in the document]
z0isch: I went with the `envKafkaBrokerAddresses :: Env.Parser Env.Error (NonEmpty BrokerAddress)` idea

library/Freckle/App/Kafka.hs
z0isch: The class based approach would force the `key` to always depend on the `value` which is probably not always the case. I'm passing `const studentId` in the current usage for example.

I went with `produceKeyedBy` et al. to keep things flexible for the future.

library/Freckle/App/Kafka.hs
```diff
@@ -0,0 +1,130 @@
...
+  logDebugNS "kafka" $ "Producing Kafka events" :# ["events" .= values]
+  view kafkaProducerPoolL >>= \case
+    NullKafkaProducerPool -> pure ()
+    KafkaProducerPool producerPool -> do
+      errors <-
+        liftIO $
+          Pool.withResource producerPool $ \producer ->
+            produceMessageBatch producer $
+              toList $
+                mkProducerRecord <$> values
```
pbrisbin, line 102: So this is an interesting interface. You're choosing to be opinionated and only support sending a batch of multiple messages (the function takes `ToJSON value => NonEmpty value` and not `ToJSON value => value`). This makes sense ergonomics-wise because the single-message case can easily be accomplished with a singleton list. I wonder if there are any other trade-offs to consider with this design. Does the underlying library also only expose `produceMessageBatch`, or does it have `produceMessage` too?

pbrisbin (approved):
The code looks good, but the style is wonky as all get-out.

Would you be able to do a quick conversion to fourmolu then bring this back overtop auto-formatted? I think this code-base was just not touched since we migrated.

library/Freckle/App/Kafka.hs
z0isch: There is also a `produceMessage` but I am not really sure about the tradeoffs involved. Both seem to just dump things on the queue that backs `librdkafka` and don't block. It looks like there is a `produceMessage'` that does have some blocking semantics if we ever need that for single messages.

The handler for Math adaptive can take multiple answers so I just exposed the `batch` variety.

z0isch:
> The code looks good, but the style is wonky as all get-out.
> 
> Would you be able to do a quick conversion to fourmolu then bring this back overtop auto-formatted? I think this code-base was just not touched since we migrated.

Will do

library/Freckle/App/Kafka.hs
pbrisbin: > I'm passing `const studentId` in the current usage for example

Sounds like you _actually_ want `produceKeyed studentId` to be the first function we have then.

But I'm very surprised `AnswerEvent` lacks a `studentId` property. Is that right? We have no need for consumers to know the student that generated an answer? I would've really expected it to, and for "always key answer events by student" to be _exactly_ the sort of thing you'd ideally define in a single place, say as an instance on `AnswerEvent`.

Even still, you could always...

```hs
data ConstantKey k a = ConstantKey k a

instance ToJSON a => ToJSON (ConstantKey k a) where
  toJSON (ConstantKey _ a) = toJSON a

instance ToKafkaPartitionKey (ConstantKey k a) where
  toKafkaPartitionKey (ConstantKey k _) = toKafkaPartitionKey a

produce $ ConstantKey studentId <$> answerEvents
```

You're totally right, of course. Not having the `Keyed` or `KeyedBy` functions available too would be a huge limitation for the library. The class-based functions would be the additional convenience for cases where that made sense.

library/Freckle/App/Kafka.hs
z0isch: 👍 That's probably a smart idea

z0isch: `produceBatchedKeyedOnAsync` is a mouthful though...

pbrisbin: Ugh, I know, not a fan.

z0isch: I think I'm just going to stick with the non-batched naming. We can adjust if we ever need the other semantics.

z0isch:
Is the nightly `CI/test` expected to be failing?

pbrisbin:
What actually dependencies nightly runs against changes ~nightly (go figure), so it's possible/common the build drifts into an un-buildable state and you don't notice until the next PR. So here we are.

I don't know what we should do ideally. In some cases we run it on a `scheduled` workflow so it fails right when it drifts, but that didn't result in better outcomes IME. We could make it non-required, but we really do want to keep things building against nightly, that's the only way you-- on wait, we explicitly don't keep `freckle-app` in stackage, so nightly failures are not particularly problematic. We could probably get away with making it not required in this repository.

Personally, I just fix it when it fails. But how easy that is varies wildly.

z0isch:
> What actual dependencies nightly runs against changes ~nightly (go figure), so it's possible/common the build drifts into an un-buildable state and you don't notice until the next PR. So here we are.
> 
> I don't know what we should do ideally. In some cases we run it on a `scheduled` workflow so it fails right when it drifts, but that didn't result in better outcomes IME. We could make it non-required, but we really do want to keep things building against nightly, that's the only way you-- on wait, we explicitly don't keep `freckle-app` in stackage, so nightly failures are not particularly problematic. We could probably get away with making it not required in this repository.
> 
> Personally, I just fix it when it fails. But how easy that is varies wildly.

Just do this in `freckle-app` it'll be easy! :trollface: 

I think the fix here may not even be possible:
```
Error: [S-4804]
       Stack failed to construct a build plan.
       
       While constructing the build plan, Stack encountered the following errors:
       
       In the dependencies for freckle-app-1.9.1.0:
           dotenv needed, but the Stack configuration has no specified version (latest matching
                  version is 0.11.0.2)
           ekg-core needed, but the Stack configuration has no specified version (latest matching
                    version is 0.1.1.7)
           yesod-core needed, but the Stack configuration has no specified version (latest matching
                      version is 1.6.24.3)
       needed since freckle-app is a build target.
```

pbrisbin:
I'm not sure why you say that's not fixable, adding those three package to `extra-deps` is a first step try.

But you can just make nightly not required, or merge around it. Don't let your dreams dreams!

PR #109, opened 2023-07-20
Add KafkaProducerPoolConfig

z0isch:
https://app.asana.com/0/1201325186562318/1204292061062064/f

We will probably want the ability to set poolSize as env variable, so this exposes a way to do that.

### DIFF
--- a/library/Freckle/App/Kafka.hs
+++ b/library/Freckle/App/Kafka.hs
@@ -1,7 +1,10 @@
+{-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE NamedFieldPuns #-}
 
 module Freckle.App.Kafka
   ( envKafkaBrokerAddresses
+  , KafkaProducerPoolConfig (..)
+  , envKafkaProducerPoolConfig
   , KafkaProducerPool (..)
   , HasKafkaProducerPool (..)
   , createKafkaProducerPool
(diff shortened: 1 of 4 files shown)


library/Freckle/App/Kafka.hs
```diff
@@ -40,37 +43,58 @@ readKafkaBrokerAddresses t = case NE.nonEmpty $ T.splitOn "," $ T.pack t of
...
   -- ^ Maximum number of resources to keep open per stripe.  The
   -- smallest acceptable value is 1.
   --
   -- Requests for resources will block if this limit is reached on a
   -- single stripe, even if other stripes have idle resources
   -- available.
+  }
+  deriving stock (Show)
+
+-- Same defaults as `Database.Persist.Sql.ConnectionPoolConfig`
```
pbrisbin, line 67:
```suggestion
-- | Same defaults as 'Database.Persist.Sql.ConnectionPoolConfig'
```

library/Freckle/App/Kafka.hs
```diff
@@ -40,37 +43,58 @@ readKafkaBrokerAddresses t = case NE.nonEmpty $ T.splitOn "," $ T.pack t of
...
+defaultKafkaProducerPoolConfig :: KafkaProducerPoolConfig
+defaultKafkaProducerPoolConfig = KafkaProducerPoolConfig 1 600 10
+
+envKafkaProducerPoolConfig
+  :: Env.Parser Env.Error KafkaProducerPoolConfig
+envKafkaProducerPoolConfig = do
+  poolSize <- Env.var Env.auto "KAFKA_PRODUCER_POOL_SIZE" $ Env.def 10
+  pure $ defaultKafkaProducerPoolConfig {kafkaProducerPoolConfigSize = poolSize}
```
pbrisbin, line 75: Whoa, we got some `ApplicativeDo` 'round here?

pbrisbin (approved):
LGTM.

Another speed-up tip: just leave this PR open and use a `git` extra-dep in backend. Only bump/merge/release when you're happily in prod.

library/Freckle/App/Kafka.hs
z0isch: `Parser` is only `Applicative`, so in order to write monadic looking parsers you have to enable it. I think we have it in some other `Parser` in `core` too.

PR #110, opened 2023-08-03
add Freckle.Test.Yesod and related modules

chris-martin:
Extracted from megarepo, organized and documented a bit better, and added the `MonadYesodExample` class.


chris-martin:
I've so far not able to get the dependencies configured with the older Stackage sets, and not sure what way to go about that.

library/Freckle/Test/Yesod/GeneralAssertions.hs
```diff
@@ -0,0 +1,40 @@
+module Freckle.Test.Yesod.GeneralAssertions
```
pbrisbin, line 1: You'll want to make sure every assertion here has `HasCallStack` all the way down (to include `liftYesodExample`, I believe) . Otherwise, the test failures will be reported as failing at the site of the underlying call(s) in this module, instead of the location of the test.

library/Freckle/Test/Yesod/JsonAssertions.hs
```diff
@@ -0,0 +1,93 @@
+module Freckle.Test.Yesod.JsonAssertions
+  ( requireJSONResponse
+  , bodyMatchesAsJson
+  , bodyArrayShouldSatisfyObjectProperties
+  , bodyDeepPermutationAsJson
+  , bodyEditEqualsAsJson
+  , bodyEqualsAsJson
```
pbrisbin, line 7: These are are legacy functions and we prefer to use the functions from `hspec-expectations-json` directly instead. Let's not bring that baggage over please.

library/Freckle/Test/Yesod/JsonAssertions.hs
```diff
@@ -0,0 +1,93 @@
...
+
+bodyArrayShouldSatisfyObjectProperties
+  :: (HasCallStack, Yesod site)
+  => [(Text, Maybe Value -> YesodExample site ())]
+  -> YesodExample site ()
+bodyArrayShouldSatisfyObjectProperties predicates = do
+  values <- NE.toList <$> getJsonBody @(NonEmpty Value)
+
+  for_ ((,) <$> values <*> predicates) $ \(value, (name, predicate)) -> do
+    predicate (value ^? key (Key.fromText name)) `orScream` T.unpack name
```
pbrisbin, line 85: As per my "these are legacy" comment, this one is interesting so I'll call it out. There is no direct replacement from `hspec-expectations-json` for it, so you might be inclined to keep it around. But I find it kind of odd and every spec I've seen use it could do something completely different, and much clearer, to express what it's asserting. :shrug:

library/Freckle/Test/Yesod/MonadYesodExample.hs
```diff
@@ -0,0 +1,25 @@
...
+  liftYesodExample :: YesodExample site a -> m a
+
+instance Yesod site => MonadYesodExample site (YesodExample site) where
+  liftYesodExample = id
+
+instance MonadYesodExample site m => MonadYesodExample site (StateT s m) where
+  liftYesodExample = lift . liftYesodExample
+
+instance MonadYesodExample site m => MonadYesodExample site (ReaderT r m) where
+  liftYesodExample = lift . liftYesodExample
```
pbrisbin, line 25: Wondering what else we should provide pass-through instances for at this time. `MaybeT`, `ExceptT`, `ResourceT`, `LoggingT`, `ValidateT`? I'm find with none, some, or all, just wanted to prompt the thought.

library/Freckle/Test/Yesod/RequestBuilder.hs
```diff
@@ -0,0 +1,75 @@
+module Freckle.Test.Yesod.RequestBuilder
+  ( -- * The RequestBuilder type
+    RequestBuilder
+
+    -- * Using a RequestBuilder
+  , request
```
pbrisbin, line 6: Should this be here or in `MakingRequests` though?

library/Freckle/Test/Yesod/StatusAssertions.hs
```diff
@@ -0,0 +1,167 @@
...
+  , postOk
+  , patchOk
+  , deleteOk
+
+    -- * Expect error
+  , getErr
+  , postErr
+  , patchErr
+  , deleteErr
+  )
```
pbrisbin, line 15: I personally don't like these, they seem like a low-value combinator that is too inflexible to be useful and just invites frustration. What if I'm using `getOk` and suddenly I need params? Where is `putOk`? What if I need headers on my `postErr`? Are we heading towards every `{method}(Ok|Err)(With(Body)(And)(Params))` combination? _We had something for that, it was called `request`. They've played us for absolute..._ anyway.

I look at something like,

```hs
getOk MyRoute
bodyDeepPermutationsAsJson whatever
```

And I just think doing the equivalent thing more directly is not much more effort, and the test is more explicit and extensible. I'm also a big fan of clearly-3-phase tests and `getOk` is mixing act/assert.

```hs
get MyRoute

statusIs 200
body <- getJsonBody
body `shouldMatchJson` whatever
```

Not really pushing back, but just offering a take in case you agree.

pbrisbin (commented):
How can I best help with the older versions problem?

chris-martin:
@pbrisbin Yes, good, I agree with pretty much all of these comments.

library/Freckle/Test/Yesod/RequestBuilder.hs
chris-martin: Maybe exported from both? If I were reading through the `RequestBuilder` module and didn't see any way to execute a `RequestBuilder`, I'd be puzzled. Likewise if I were looking through the `MakingRequests` module to find out what utilities were available for making requests, I'd want to be pointed to the `RequestBuilder` tool.

library/Freckle/Test/Yesod/GeneralAssertions.hs
```diff
@@ -0,0 +1,40 @@
...
+  , assertNotEq
+  )
+where
+
+import Freckle.App.Prelude
+import Freckle.Test.Yesod.MonadYesodExample
+
+import qualified Yesod.Test
+
+-- | Asserts that the two given values are equal
```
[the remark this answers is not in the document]
chris-martin: I didn't check to see if we actually use these; looks like we don't.

library/Freckle/Test/Yesod/JsonAssertions.hs
chris-martin: https://github.com/freckle/megarepo/pull/29168 removes all of these, except `bodyArrayShouldSatisfyObjectProperties` which will require a tad more thought.

library/Freckle/Test/Yesod/JsonAssertions.hs
chris-martin: https://github.com/freckle/megarepo/pull/29178 removes `bodyArrayShouldSatisfyObjectProperties`, it's a lot simpler without it.

chris-martin:
> How can I best help with the older versions problem?

"Copy the build constraints from `bcp47` configurations" is what I needed to hear.

The only configuration that still doesn't build is nightly, due to https://github.com/lexi-lambda/monad-validate/pull/1

PR #112, opened 2023-08-09
Add Nix flake

stackptr:
Follows example in other repos to allow building and testing of `freckle-app` via `nix develop`. There are no tooling changes for non-Nix users.


pbrisbin (approved):
> There are no tooling changes for non-Nix users

Love it.

stackptr:
@chris-martin when using `nix develop` last night, it took quite a while (maybe close to two hours?) before I had a Nix shell. I think this is because this uses a different set of GHC packages, as megarepo and other repos use `ghc927` while this uses `ghc926` (based on the `stack.yaml` resolver), so a bunch of stuff had to be built from scratch. Possibly this is more performant on a later stack resolver?

chris-martin:
@stackptr I think the biggest reason it took hours to build is probably because we've got haskell-language-server in there, and since it's overridden to use a particular patch version of GHC, it can't pull from the cache. It helps to keep things on the same versions if you copy `flake.lock` from another repo instead of letting Nix generate it (don't know which you did) but of course changing the GHC version is going to mean rebuilding a lot.

I think we're really going to adopt a service so we can have our own binary cache sooner than later.

chris-martin:
I think there are a few options now. The one I'm familiar with is [Cachix](https://www.cachix.org/):

- We'd create a cachix account named e.g. `freckle`. 
- Everybody installs the `cachix` command-line utility, and runs `cachix use freckle` to add our cache to their local list of trusted remote caches that Nix will try when building anything.
- There's, uh, some kind of secret key we'd need to share around to make each person able to upload stuff to the cache
- To upload to the cache, you pipe a list of nix store paths into the `cachix upload` command, e.g. `nix build [whatever] --no-link --print-out-paths | cachix upload freckle`

lafrenierejm:
> There's, uh, some kind of secret key we'd need to share around to make each person able to upload stuff to the cache

Why would individuals need to upload directly to the cache? Would it be sufficient to give devs read-only access and let CI upload on each build?

chris-martin:
> let CI upload

Yep, that probably makes more sense!

PR #113, opened 2023-08-11
Add Kafka consumer module

stackptr:
Adds `Freckle.App.Kafka.Consumer` to support applications in setting up a Kafka consumer that polls for messages sent by a Kafka producer (see #106). `envKafkaConsumerConfig` reads the environment for necessary variables to set up the consumer, including the group id and topics to poll. `withKafkaConsumer` uses the configuration to create the consumer and `runConsumer` accepts a timeout and callback to run when a message is received:

```haskell
-- App with instances HasLogger, HasKafkaConsumer

timeoutMs :: Int
timeoutMs = 1000

loadApp :: (App -> IO a) -> IO a
loadApp f = do
  logConfig <- parse id LoggingEnv.parser
  kafkaConfig <- parse id envKafkaConsumerConfig
  logger <- newLogger logConfig
  withKafkaConsumer kafkaConfig $ \consumer ->
    f $ App logger consumer

main :: IO ()
main =
  runApp loadApp $ runConsumer @_ @_ @Value timeoutMs $
    \message -> logInfo $ "Message receieved" :# ["message" .= message]
```

Also creates a general `timeout` function from existing logic that reads in a timeout for use in Postgres. This is convenient to use in setting up the polling interval in Kafka consumers.

Planned for use in https://github.com/freckle/progress/pull/42


library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -0,0 +1,166 @@
...
+  -- ^ The consumer group id to which the consumer belongs.
+  --
+  -- This is the `group.id` Kafka consumer configuration property.
+  , kafkaConsumerConfigTopics :: NonEmpty TopicName
+  -- ^ The list of topic names that for which the consumer will subscribe.
+  , kafkaConsumerConfigOffsetReset :: OffsetReset
+  -- ^ The offset reset parameter used when there is no initial offset in Kafka.
+  --
+  -- This is the `auto.offset.reset` Kafka consumer configuration property.
+  , kafkaConsumerConfigExtraSubscriptionProps :: Map Text Text
```
z0isch, line 49: I am assuming this is to allow configuration of the underlying `Subscription`? Maybe a mention of `Kafka.Consumer` would be nice here.

library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -0,0 +1,166 @@
+{-# LANGUAGE ApplicativeDo #-}
+
+module Freckle.App.Kafka.Consumer
```
z0isch, line 3: I wonder if we need to rename `Freckle.App.Kafka` to `Freckle.App.Kafka.Producer`?

library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -0,0 +1,166 @@
...
+  => Timeout
+  -> (a -> m ())
+  -> m ()
+runConsumer pollTimeout onMessage = void $ Immortal.create $ \thread -> Immortal.onUnexpectedFinish thread handleException $ do
+  consumer <- view kafkaConsumerL
+  eMessage <-
+    pollMessage consumer $ Kafka.Timeout $ timeoutMs pollTimeout
+  case eMessage of
+    Left (KafkaResponseError RdKafkaRespErrTimedOut) -> logInfo $ "Polling timeout"
+    Left err -> logError $ "Error polling for message from Kafka" :# ["error" .= show err]
```
z0isch, line 152: Any reason in particular to split these 2 errors up?

library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -0,0 +1,166 @@
...
+timeoutMs = \case
+  TimeoutSeconds s -> s * 1000
+  TimeoutMilliseconds ms -> ms
+
+runConsumer
+  :: ( MonadUnliftIO m
+     , MonadReader env m
+     , MonadLogger m
+     , HasKafkaConsumer env
+     , FromJSON a
```
z0isch, line 141: I wonder how often we would subscribe to more than one topic which happen to have the same underlying `FromJSON a`? 

It seems like we may want to take something like `TopicName -> Value -> a -> m ()` instead of just `a -> m ()` but that feels like trying to support a need we just might not have. 

I wonder if we should make `kafkaConsumerConfigTopics :: TopicName` instead?

library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -0,0 +1,166 @@
...
+    Left (KafkaResponseError RdKafkaRespErrTimedOut) -> logInfo $ "Polling timeout"
+    Left err -> logError $ "Error polling for message from Kafka" :# ["error" .= show err]
+    Right (ConsumerRecord {..}) -> for_ crValue $ \bs -> do
+      case eitherDecodeStrict bs of
+        Left err -> logError $ "Could not decode message value" :# ["error" .= err]
+        Right a -> onMessage a
+  maybe
+    (logInfo "Committed offsets")
+    (\err -> logError $ "Error committing offsets" :# ["error" .= show err])
+    =<< commitAllOffsets OffsetCommit consumer
```
z0isch, line 160: Does having this `commitAllOffsets` outside of the `case eMessage` imply that we skip messages that have some kind of error?

library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -0,0 +1,166 @@
...
+ where
+  handleException = \case
```
z0isch, line 162: I guess exceptions caught by this handler would mostly mean that we most likely retry the message?

library/Freckle/App/Kafka/Consumer.hs
pbrisbin: Sounds right, and then `Freckle.App.Kafka` could re-export both.

library/Freckle/App/Kafka/Consumer.hs
pbrisbin: The polling timeout error is expected, frequently. Every time there is no event withing some period it will throw that. I'd probably DEBUG, FWIW.

z0isch: Aha! Yeah I think debug is the way to go then

library/Freckle/App/Kafka/Consumer.hs
pbrisbin: > we skip messages that have some kind of error

I think we want that, right? Otherwise, any error will cause our consumers to stop making any progress. Not ideal when that happens at 4am for a reason you don't care about.

I could see this being configurable behavior, but I think we want the default to be lenient.

z0isch: I think this is a super tricky subject and might prefer to have this not be lenient by default. Seems like we could take an error handling function and still offer this behavior conveniently 

```hs
data CommitOffset = CommitOffset | DoNotCommitOffset
onError :: KafkaError -> m CommitOffset
alwaysCommit = const CommitOffset
```

library/Freckle/App/Kafka/Consumer.hs
stackptr: Kind of; `hw-kafka-client` has no documentation on these props -- you have to use the Kafka Consumer documentation instead. I was hoping this is legible by other record properties above which reference `bootstrap.servers`, `group.id`, and `auto.offset.reset`.

library/Freckle/App/Kafka/Consumer.hs
stackptr: I'm not sure if there is a change being suggested, but I want to clarify that `runConsumer` uses the `Subscription` that was setup in `withKafkaConsumer -> KafkaConsumerConfig -> ...` and so the first example above is actually:

```haskell
config :: KafkaConsumerConfig
config = KafkaConsumerConfig brokers groupId topics Earliest mempty
 where
  topics = NE.fromList [topicA, topicB]

loadApp :: (App -> IO a) -> IO a
loadApp f = do
  -- ...
  withKafkaConsumer config $ \consumer -> f $ App consumer

runConsumer $ \messageFromAorB -> do
```

I'm not sure how the second example would work as I think it entails having two different consumers. The first example is what I believe follows from the library because of how the `Subscription` cannot be changed/modified in a consumer that is setup, but the second example is how I originally thought the consumer API would work.

z0isch: My suggestion was to make `kafkaConsumerConfigTopics :: TopicName` instead as it would be pretty funky to support listening to more than one topic IMO.

pbrisbin: Just to clarify: my comment was just agreeing with AJ and showing examples in a simplified syntax. The intent was to show that passing `[topicA, topicB]` to a single consume (how the PR could be configured now) was not preferred or useful, and passing a single topic to multiple consumes (how AJ suggests the config be changed) is how we would most likely want to achieve a multi-topic app anyway.

library/Freckle/App/Kafka/Consumer.hs
pbrisbin: Yeah you're right it's very tricky and I agree it should be configurable. I don't feel strongly on the default behavior, you can decide with Corey.

library/Freckle/App/Kafka/Consumer.hs
stackptr: I am not clear on the utility of a single message handler processing across multiple topics, but I think this should align with the upstream library until we have a concrete reason to diverge and do our own thing.

library/Freckle/App/Kafka/Consumer.hs
stackptr: I don't think it is likely since the offset is committed after processing the message, which I'd think is where exceptions could mostly occur.

library/Freckle/App/Kafka/Consumer.hs
stackptr: Fair enough!

library/Freckle/App/Kafka/Consumer.hs
stackptr: There was some back and forth about this over Slack. d4a8830 uses `commitOffsetMessage`, making the default logic to commit an offset only when a message has been successfully parsed and processed. There might be some value in allowing a client to configure the offset commit to occur if parsing and/or processing fails, but there doesn't seem to be any value in allowing for the commit to occur on `KafkaError`.

z0isch (approved):
Looks good to me. I'm curious to get this out in the wild and seeing what sort of `KafkaError`s we get in practice and what that should mean for committing offsets.

PR #120, opened 2023-09-22
Full support for OpenTelemetry tracing, use it instead of XRay in `runDB`

pbrisbin:
This was spiked, implemented, used in platform-example-app, then shelved
because megarepo wasn't ready for it. I got nervous having the code live in
a branch and devised a way to land it in `main` so we could use it from Platform
Apps, but still be able to keep megarepo on XRay.

The CHANGELOG sums up well where we landed:

This replaces our bespoke, XRay-based `MonadTracer` class to be that of
hs-opentelemetry. This means `Freckle.App.Database.runDB` now traces using
OpenTelemetry instead of XRay. A compilation error will indicate a missing
`HasTracer` instance, which you should implement based on the
`Freckle.App.OpenTelemetry` module documentation.

Clients who wish to remain on XRay can still upgrade, but must adjust their
imports so they use `Freckle.App.Database.XRay` instead. The
`Freckle.App.Database` module is re-exported through `Freckle.App` and
`Freckle.App.Test`, so there may be some additional `hiding` required. A
compilation error indicating a missing `HasTracer` or `MonadTracer` instance
means your imports are not yet correct.

---

I've already tested using this from megarepo (staying on XRay) and progress
(opting in for tracing), so I'm relatively confident the `.XRay` module plan works.


pbrisbin:
> The Freckle.App.Database module is re-exported through Freckle.App and Freckle.App.Test, so there may be some additional hiding required

This is actually a smell, IMO, that I have to deal with this in our own clients, so I'm going to remove the re-exports and see if things work out better.

It _should_ mean adding `import Freckle.App.Database.XRay` now, but then being able to uniformly `s/.XRay//` later to move onto OTel, which is better than having to remove and/or undo `hiding`. :crossed_fingers:

library/Network/Wai/Middleware/OpenTelemetry.hs
```diff
@@ -0,0 +1,66 @@
...
+          [("X-Datadog-Trace-Id", traceIdIntBS)]
+          app
+          request
+          respond
+
+-- | <https://github.com/iand675/hs-opentelemetry/blob/1f0328eb59fec2a97aec7ef98fe4f1e0d5c8f2ac/propagators/datadog/src/OpenTelemetry/Propagator/Datadog.hs#L103>
+ddTraceId :: TraceId -> Word64
+ddTraceId traceId = indexByteArrayNbo (ByteArray a) 1
+ where
+  !(SBI.SBS a) = SBI.toShort $ traceIdBytes traceId
```
chris-martin, line 55: Well this is all very upsetting

library/Freckle/App/OpenTelemetry.hs
```diff
@@ -1,18 +1,67 @@
...
-  getVaultData :: m (Maybe XRayVaultData)
+withTracerProvider :: MonadUnliftIO m => (TracerProvider -> m a) -> m a
+withTracerProvider =
+  bracket
+    (liftIO initializeGlobalTracerProvider)
+    (liftIO . shutdownTracerProvider)
 
-instance MonadTracer (HandlerFor app) where
-  getVaultData = vaultDataFromRequest <$> waiRequest
+getCurrentTraceId :: MonadIO m => m (Maybe (TraceId, SpanId))
```
chris-martin, line 64: Minor naming quibble: I you don't want to go so wordy as `getCurrentTraceAndSpanId`, why not `getCurrentSpanId` since that's the more specific of the two pieces of information here?

chris-martin (approved):
Feel like I'm missing enough xray/otel knowledge to give substantive feedback here, but I think it all makes enough sense.

library/Freckle/App/OpenTelemetry.hs
pbrisbin: I now forget why this returns both. It's only used for the trace-id part anyway, which is probably where the naming came from. Let me clean this up a bit.

PR #121, opened 2023-10-02
Ensure Kafka consumers include trace_id in log context

pbrisbin:
### [Centralize adding TraceId to the logging context](https://github.com/freckle/freckle-app/pull/121/commits/34671b2cf78ffdd0aec3a68c8d7db5f08e2bbe1f)
[34671b2](https://github.com/freckle/freckle-app/pull/121/commits/34671b2cf78ffdd0aec3a68c8d7db5f08e2bbe1f)

This commit adds some functions to OpenTelemetry for working with the
current trace-id as converted to Datadog's format and for the purposes
of setting the log context. It's then re-used in the Middleware from
which it was extracted.

This will allow us to do the same thing in other persistent-process
runners, such as Kafka consumers, just like we do in APIs through the
middleware.

### [Wrap runConsumer in withTraceIdContext](https://github.com/freckle/freckle-app/pull/121/commits/98da8d6d5173937b0807920f7431718172f433b4)
[98da8d6](https://github.com/freckle/freckle-app/pull/121/commits/98da8d6d5173937b0807920f7431718172f433b4)

This will ensure consumer logs all contain the current `trace_id`.

### [Version bump](https://github.com/freckle/freckle-app/pull/121/commits/e3a2d82dc2d651e2e67d3a9e895a76167ca55f43)

### DIFF
--- a/library/Network/Wai/Middleware/OpenTelemetry.hs
+++ b/library/Network/Wai/Middleware/OpenTelemetry.hs
@@ -5,15 +5,11 @@ module Network.Wai.Middleware.OpenTelemetry
 
 import Freckle.App.Prelude
 
-import Blammo.Logging (withThreadContext, (.=))
 import qualified Data.ByteString.Char8 as BS8
 import Freckle.App.OpenTelemetry
 import Network.Wai
 import Network.Wai.Middleware.AddHeaders
 import qualified OpenTelemetry.Instrumentation.Wai as Trace
-import OpenTelemetry.Propagator.Datadog
-  ( convertOpenTelemetryTraceIdToDatadogTraceId
-  )
 
 newOpenTelemetryWaiMiddleware :: IO Middleware
 newOpenTelemetryWaiMiddleware = do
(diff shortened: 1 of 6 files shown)


library/Network/Wai/Middleware/OpenTelemetry.hs
```diff
@@ -27,18 +23,13 @@ newOpenTelemetryWaiMiddleware = do
...
 
-      withThreadContext ["trace_id" .= traceIdInt] $
-        addHeaders
-          [("X-Datadog-Trace-Id", traceIdIntBS)]
-          app
-          request
-          respond
+addTraceIdHeader :: Middleware
+addTraceIdHeader app request respond = do
+  mTraceId <- getCurrentTraceIdAsDatadog
```
z0isch, line 31: Did you consider a `withTraceIdContext :: (MonadIO m, MonadMask m) => (Maybe Word64 -> m a) -> m a` to avoid the awkward `getCurrentTraceIdAsDatadog` inside of `withTraceIdContext`?

z0isch (approved):

library/Network/Wai/Middleware/OpenTelemetry.hs
pbrisbin: I did. It seemed to make other, more general use-cases more awkward to make the middleware use-case less awkward, which I thought was a poor trade-off.

pbrisbin: Perhaps something like...

```hs
withTraceIdContext :: m a -> m a
withTraceIdContext f = withTraceIdContext' $ \_ -> f

withTraceIdContext' :: (Maybe Word64 -> m a) -> m a
withTraceIdContext'
```

I just wonder what value this brings to make for the fact now we have to talk about naming this stuff.

pbrisbin: Since I can extract that later without braking compatibility, I'm going to merge as is, but we can keep talking.

z0isch: Meh, I think it's going to be ugly either way. I'd probably feel less confused seeing `withTraceIdContext $ \_ -> do ...` then the `withTraceIdContext $ do { mTraceId  <- withTraceIdContext; ...}` but I don't feel strongly at all.

pbrisbin: That's the rub though. `withTraceIdContext $ \_` implies `(Maybe TraceId -> ...` not `(Maybe Word64`. So you either have to give it a name that says it's Datadog specific, or you need the caller to call a different (Datadogly-named) function to do the `TraceId -> Word64` conversion themselves. None of which I enjoyed very much.

PR #123, opened 2023-10-17
upstream Freckle.Memcached from megarepo/backend/entities

chris-martin:
I'd like to be able to use some of this from megarepo/backend/core; it's currently in `entities`. Could move it to `core`, but might as well push it all the way out? There's a commit message indicating there was intent to upstream it; looks like it just never happened.


pbrisbin (approved):
Can confirm doing this was always debt: https://app.asana.com/0/399653121150251/1203420543603647/f

PR #125, opened 2023-10-25
Add tracing to Memcached and Kafka operations

pbrisbin:
### [Fix import order](https://github.com/freckle/freckle-app/pull/125/commits/b82b2d7f2df83f23f84c6b56599023fdfd9e3782)
[b82b2d7](https://github.com/freckle/freckle-app/pull/125/commits/b82b2d7f2df83f23f84c6b56599023fdfd9e3782)

### [Add kind-related helpers for SpanArguments](https://github.com/freckle/freckle-app/pull/125/commits/02c522f9ee2dd44c67e41d7bce04d1047155f374)
[02c522f](https://github.com/freckle/freckle-app/pull/125/commits/02c522f9ee2dd44c67e41d7bce04d1047155f374)

### [Use trace kind Client for runDB span](https://github.com/freckle/freckle-app/pull/125/commits/6ee57d5c5d1c913bf63a3c2b990c6337eb43e439)
[6ee57d5](https://github.com/freckle/freckle-app/pull/125/commits/6ee57d5c5d1c913bf63a3c2b990c6337eb43e439)

### [Add tracing to Memcached operations](https://github.com/freckle/freckle-app/pull/125/commits/b4fe4109ac1cd0d01c77aaa051f5b95407cfa9ae)
[b4fe410](https://github.com/freckle/freckle-app/pull/125/commits/b4fe4109ac1cd0d01c77aaa051f5b95407cfa9ae)

### [Add tracing to Kafka.Producer operations](https://github.com/freckle/freckle-app/pull/125/commits/59f17ae01223924560e3aeec454f08234f3226b2)
[59f17ae](https://github.com/freckle/freckle-app/pull/125/commits/59f17ae01223924560e3aeec454f08234f3226b2)

This only adds topic name to the attributes, since how to deal with
multiple values and a keying function would be somewhat arbitrary. We'll
leave out that behavior for now and possibly add something later when we
start to use this and decide what we'd want.

### [Add tracing to Kafka.Consumer operations](https://github.com/freckle/freckle-app/pull/125/commits/d8ad053059dac0b398ccffc8b3449e2f6b75f2d4)
[d8ad053](https://github.com/freckle/freckle-app/pull/125/commits/d8ad053059dac0b398ccffc8b3449e2f6b75f2d4)

This adds an outer span, and inner spans for decoding and handling the
message.

In order for spans to be correctly marked as error with exception
details, you need to "throw through" the `inSpan` function, so I had to
[...]

PR #128, opened 2023-10-26
add MonadSql, MonadSqlTx

chris-martin:
As discussed in https://github.com/freckle/megarepo/pull/30864


chris-martin:
Okay, I think this part is ready.

library/Freckle/App.hs
```diff
@@ -251,6 +252,12 @@ instance (Monad m, HasTracer app) => MonadTracer (AppT app m) where
 instance Applicative m => XRay.MonadTracer (AppT app m) where
   getVaultData = pure Nothing
 
+instance
+  (MonadUnliftIO m, HasSqlPool app, HasStatsClient app)
+  => MonadSqlTx (ReaderT SqlBackend (AppT app m)) (AppT app m)
+  where
+  runSqlTx = XRay.runDB
```
pbrisbin, line 259: `XRay` is legacy, and only exists because megarepo can't migrate from XRay to Datadog until November 1 (when we get an APM trial in place). You want you use `Freckle.App.Database.runDB` for this -- I don't _think_ that'll be any kind of issue (for example, I don't think need to duplicate anything like we do with `MonadTracer`).

library/Freckle/App/Database.hs
```diff
@@ -60,6 +74,32 @@ import UnliftIO.Exception (displayException)
...
+--   @db@ has access to a connection.
+class (MonadSqlBackend db, Monad m) => MonadSqlTx db m | m -> db where
+  -- | Runs the action in a SQL transaction
+  runSqlTx :: HasCallStack => db a -> m a
+
+class HasSqlBackend a where
+  getSqlBackend :: a -> SqlBackend
+
+instance HasSqlBackend SqlBackend where
+  getSqlBackend = id
```
pbrisbin, line 101: My spidey sense is tingling a bit that we have/need this but also `HasSqlPool`. Seems to be very overlapping concerns (get a connection so we can `runDB`).

Assuming there's a valid reason for both, should we include a mirroring `HasSqlBackend (HandlerData child site)` too?

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.10.2.0...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.10.3.0...main)
+
+## [v1.10.3.0](https://github.com/freckle/freckle-app/compare/v1.10.2.0...v1.10.3.0)
+
+- Add classes `MonadSqlBackend` and `MonadSqlTx` for being quite nonspecific about the
+  context in which you interact with a SQL database.
```
pbrisbin, line 6: So what happens to the following expressions now?

```hs
foos :: (MonadUnliftIO m, MonadTracer m, MonadReader env m, HasSqlPool env, HasStatsClient env) => m [Foo]
foos = runDB fetchFoos

fetchFoos :: MonadIO m => SqlPersistT m [Foo]
fetchFoos = selectList [...] [...]
```

Will it look like this?

```hs
foos :: MonadSqlTx m => m [Foo]
foos = runDB $ liftSql fetchFoos
```

Or I guess this?

```hs
foos :: MonadSqlTx m => m [Foo]
foos = runDB fetchFoos

fetchFoos :: MonadSqlBackend m => m [Foo]
fetchFoos = liftSql $ selectList [...] [...]
```

And can you expand a bit on why that's useful?

chris-martin: At the expression level, yes, the latter.

```haskell
foos = runDB fetchFoos
fetchFoos = liftSql $ selectList [...] [...]
```

Unless we wanted to make our own polymorphic wrapper for persistent, in which case nothing at the expression level would change at all.

Types:

```haskell
foos :: (MonadSqlTx db m, MonadIO m, MonadIO db)
fetchFoos :: (MonadSqlBackend m, MonadIO m) => m [Foo]
```

(Eh, maybe `MonadIO` superclasses would be better.)

The purpose is for specs to be able to define a monad whose instances may be fakes. The generalization is useful when `db` isn't `SqlPersistT m`, which will be the case any time `GraphulaTestT` is in the stack.

pbrisbin: > monad whose instances may be fakes
> ...
> without having to explicitly lift into the test monad

Are these two orthogonal purposes or are they related? I ask because I just don't ever see us doing the first thing, so I don't know if the second thing is independently useful.

> Unless we wanted to make our own polymorphic wrapper for persistent

I doubt we'll do that right?

So is it fair to say this change is kind of a syntactic wash?

```diff
- lift $ runDB selectList
+ runDB $ liftSql selectList
```

> (Eh, maybe `MonadIO` superclasses would be better.)

Yeah, I think we mentioned somewhere else that `MonadSqlTx` could have a `MonadUnliftIO` superclass right? Any reasonable transaction is going to require a bracket.

But I think this simplification in the constraints, and what you said somewhere before about clients that don't care about expression or tracing don't have to mention it, is the _actual_ purpose/utility of this, right?

> for when db isn't `SqlPersistT m`, which will be the case any time `GraphulaTestT` is in the stack.

I find this so irrationally annoying :joy:

library/Freckle/App/Database.hs
chris-martin: They seem very different to me, being able to get a pool versus being able to get a single connection.

I don't see what the purpose of `HasSqlBackend (HandlerData child site)` would be. Won't the `site` always have a pool, not a single connection?

pbrisbin: Yeah, I guess they are practically different and I was conflating them because to me it's conceptually the same with the Pool being an implementation details.

```hs
runDB :: (MonadUnliftIO m, MonadReader env m, HasSqlPool env) => ...
runDB b = do
  p <- view sqlPoolL
  Pool.with p $ \b ->
    Persist.actuallyRunSqlQuery b q

runDB :: (MonadUnliftIO m, MonadReader env m, HasSqlBackend env) => ...
runDB b = do
  p <- getSqlBackendM
  Persist.actuallyRunSqlQuery b q
```

You can even imagine

```hs
class (MonadReader env m, HasSqlPool env) => MonadSqlBackend m where
  getSqlBackendM = do
    p <- view sqlPool
    Pool.with p pure
```

I mentioned spidey sense in my comment because I didn't have an actual concrete question or concern, these ideas were just floating around and thought there might be something there.

I think I've set to see us actually define a `HasSqlBackend` instead for anything other than `ReaderT SqlBackend` and I've yet to see us actually use `getSqlBackendM` anywhere -- so I'm still struggling to bring this into the realm of the practical.

pbrisbin (changes requested):
I could spend all day trying to learn more about this stuff, but the only thing I'm blocking on is removing `Xray.runDB` from `Freckle.App`. Once that's done I'll approve and we can move on.

chris-martin:
I used the XRay version because that's what we're actually using, with the intent to change it later. Went ahead and changed it to otel, but it means this project is on hold since I can't actually use the instances in megarepo yet.

pbrisbin:
> since I can't actually use these instances in megarepo yet

Ah, well that is annoying. I had made the cut-over to make `freckle-app` OTel with XRay opt-in, so yeah, violating that does put it on hold. The good news is I plan to merge that in megarepo on Nov 1. Then we can get this and #125 in, which is also blocked by that. :crossed_fingers:

library/Freckle/App.hs
```diff
@@ -248,8 +248,14 @@ instance PrimMonad m => PrimMonad (AppT app m) where
 instance (Monad m, HasTracer app) => MonadTracer (AppT app m) where
   getTracer = view tracerL
 
-instance Applicative m => XRay.MonadTracer (AppT app m) where
-  getVaultData = pure Nothing
+-- instance Applicative m => HasTracer (AppT app m) where
+--   getVaultData = pure Nothing
```
pbrisbin, line 252: I would revert this. The `XRay.MonadTracer` instance still works (and is needed) for things opted into `XRay`; they just can't use `MonadSqlTx` yet. And the commented `HasTracer` instance isn't valid even if uncommented.

library/Freckle/App/Test.hs
```diff
@@ -122,8 +126,11 @@ instance HasLogger app => Example (AppExample app a) where
 instance HasTracer app => MonadTracer (AppExample app) where
   getTracer = view tracerL
 
-instance XRay.MonadTracer (AppExample app) where
-  getVaultData = pure Nothing
```
pbrisbin, line 126: You'll want to keep this. Anything still opting into `XRay` can function even with your changes, they just have to use `Database.XRay.runDB` and can't get the `MonadSql*` goodies. Though I know it practically doesn't matter, as we're really the only user, I see no real downside to not to saving removing this for when we remove everything else, in a breaking-change version bump.

pbrisbin (changes requested):
I apologize for the confusion of the `.XRay` stuff, and generating extra work for something that was temporary and very close to gone, but I think we want to keep that working (keep those instances I commented on) across this change so we can remove it altogether as its own atomic update later. Or you (or I, if you want) could remove it first, then rebase your work on top, since this stuff will be blocked until we're done with it anyway.

library/Freckle/App.hs
chris-martin: oof, I've further confused this because this isn't even something I meant to remove.

library/Freckle/App/Database.hs
```diff
@@ -60,6 +75,32 @@ import UnliftIO.Exception (displayException)
...
+-- | A monadic context in which a SQL backend is available
+--   for running database queries
+class MonadIO m => MonadSqlBackend m where
+  getSqlBackendM :: m SqlBackend
+
+instance (HasSqlBackend r, MonadIO m) => MonadSqlBackend (ReaderT r m) where
+  getSqlBackendM = asks getSqlBackend
+
+-- | Can be used e.g. to generalized from 'SqlPersistT' to 'MonadSqlBackend'
+liftSql :: MonadSqlBackend m => ReaderT SqlBackend m a -> m a
```
pbrisbin, line 87: Take it or leave it :)

```suggestion
-- | Generalize 'SqlPersistT' to 'MonadSqlBackend'
liftSql :: MonadSqlBackend m => SqlPersistT m a -> m a
```

library/Freckle/App/Database.hs
```diff
@@ -60,6 +75,32 @@ import UnliftIO.Exception (displayException)
...
+liftSql (ReaderT f) = getSqlBackendM >>= f
+
+-- | The constraint @'MonadSqlTx' db m@ indicates that @m@ is a monadic
+--   context that can run @db@ actions within a SQL transaction.
+--   Typically this means that @m@ has access to a connection pool and
+--   @db@ has access to a connection.
```
pbrisbin, line 93: I still trip over all this, so here's another take-it-or-leave it rewrite for my own benefit:

```suggestion
-- | The constraint @'MonadSqlTx' db m@ indicates that @m@ is a monadic
--   context that can run @db@ actions, usually as a SQL transaction.
--   Typically, this means that @db@ needs a connection and @m@ can
--   provide one, e.g. from a connection pool.
```

pbrisbin (approved):

PR #129, opened 2023-11-02
Run tests with JUnit from env vars, if enabled

stackptr:
Uses `JUNIT_ENABLED` and other env vars to trigger and configure JUnit formatting when running tests, removing the legacy `CIRCLECI` env var. JUnit configuration is overridden to set the output path to `/tmp/junit` (existing behavior) and the source path prefix of the report, which is helpful in collecting test results in a monorepo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.11.0.0...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.10.4.0...main)
 
-## [v1.11.0.0](https://github.com/freckle/freckle-app/compare/v1.10.3.0...v1.11.0.0)
+## [v1.10.4.0](https://github.com/freckle/freckle-app/compare/v1.10.3.0...v1.10.4.0)
 
 - Use JUnit formatting for test output when `JUNIT_ENABLED` environment
   variable is set.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,14 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.10.3.0...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.11.0.0...main)
+
+## [v1.11.0.0](https://github.com/freckle/freckle-app/compare/v1.10.3.0...v1.11.0.0)
+
+- Use JUnit formatting for test output when `JUNIT_ENABLED` environment
+  variable is set.
+- Use environment variables to [configure JUnit][junit-env]; always prefix
+  source paths with test suite `name` and output tests to `/tmp/junit`.
+- Remove using `CIRCLECI` environment variable.
+
+[junit-env]: https://github.com/freckle/hspec-junit-formatter/blob/main/library/Test/Hspec/JUnit/Config/Env.hs
 
 ## [v1.10.3.0](https://github.com/freckle/freckle-app/compare/v1.10.2.0...v1.10.3.0)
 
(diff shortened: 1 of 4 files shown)


stackptr:
I couldn't figure out how to use `--format junit` when using `configWithJUnitAvailable`. My local testing indicated that a sufficient version of `hspec-core` is used, yet the test argument flag is unrecognized. As a result, I added [`90bbcf6`][90bbcf6] to do a modified version of `hspecJUnitWith` from hspec-junit-formatter, configuring JUnit from env vars and then applying overrides for output file and source path prefix.

Output is unchanged when running tests locally with the usual test arguments, but now a JUnit report with the fixed source paths is generated when invoked with `JUNIT_ENABLED=1`.

[90bbcf6]: https://github.com/freckle/freckle-app/pull/129/commits/90bbcf694df60894959b189850e7d617218e9029

library/Freckle/App/Test/Hspec/Runner.hs
```diff
@@ -66,18 +65,21 @@ runWith config name spec = do
...
   noConcurrency x = x {configConcurrentJobs = Just 1}
 
-runJUnitSpec :: Spec -> (FilePath, String) -> Config -> IO Summary
-runJUnitSpec spec (path, name) config =
-  spec `runSpec` configWithJUnit junitConfig config
+runWithJUnit :: Spec -> (FilePath, String) -> Config -> IO Summary
+runWithJUnit spec (path, name) config = do
+  junitEnabled <- envJUnitEnabled
+  baseJUnitConfig <- envJUnitConfig
+  let modify = if junitEnabled then configWithJUnit $ withOverride baseJUnitConfig else id
```
pbrisbin, line 76: Are you doing this because you can't make `-f junit` work? I would've expected something like,

```hs
junitConfig <- withOverride <$> envJUnitConfig
spec `runSpec` configWithJunitAvailable junitConfig config
```

pbrisbin (commented):
> like a fine default since this test runner is only used in monorepos

Is it a default that can be changed, or just the way it always works?

freckle-app is tricky because, while we pretty much know only we use it, it is open source and could be used "in the wild" so, when it's not too cumbersome, we do try and keep it less opinionated on stuff like that like that. I, at least, find it results in a better package for on our future uses too.

library/Freckle/App/Test/Hspec/Runner.hs
pbrisbin: Just saw you other comment. I'd really like to understand why this is not working, but I get it if you just want to go ahead this way anyway.

stackptr:
> > like a fine default since this test runner is only used in monorepos
> 
> Is it a default that can be changed, or just the way it always works?
> 
> freckle-app is tricky because, while we pretty much know only we use it, it is open source and could be used "in the wild" so, when it's not too cumbersome, we do try and keep it less opinionated on stuff like that like that. I, at least, find it results in a better package for on our future uses too.

I agree in principle. In the case of this test runner module, using `name` to output distinct test results (either as JUnit reports or in the logging) feels rather opinionated already, and using it to set source prefixes a lateral move.

CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.10.3.0...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.11.0.0...main)
+
+## [v1.11.0.0](https://github.com/freckle/freckle-app/compare/v1.10.3.0...v1.11.0.0)
```
pbrisbin, line 3: I could see this not being a major bump, FWIW.

pbrisbin (approved):

CHANGELOG.md
stackptr: I was on the fence and ended up with a major bump because of dropping using the `CIRCLECI` env var. I'll change it to a minor bump since using that env var is arguably an implementation detail, not a public API

PR #130, opened 2023-11-09
Add Freckle.App.Random

chris-martin:
For the sake of [Let Graphula tests select more than one random entity](https://app.asana.com/0/1205102828747541/1205920572323612). With this utility available, I think it should be easy enough enough to extend `Freckle.Test.Graphula.selectRandom` to randomly generating a set of row numbers and then select where `row_number()` is in the set.

### DIFF
--- a/library/Freckle/App/Random.hs
+++ b/library/Freckle/App/Random.hs
@@ -3,6 +3,7 @@
 module Freckle.App.Random
   ( smallRandomSubsetOfLargeIntegerRange
   , Range (..)
+  , NonEmptyRange (..)
   , inclusiveRange
   ) where
 
@@ -17,10 +18,16 @@ import Numeric.Natural (Natural)
 
 import qualified Data.Set as Set
 
--- | A contiguous range of integers
-data Range i = Range
+-- | A possibly-empty contiguous range of integers
+data Range i
+  = RangeEmpty
+  | RangeNonEmpty (NonEmptyRange i)
+
+-- | A nonempty contiguous range of integers
+data NonEmptyRange i = NonEmptyRange
   { inclusiveMinBound :: i
-  , size :: Natural
+  , offset :: Natural
+  -- ^ The size of the range minus one
   }
 
 inclusiveRange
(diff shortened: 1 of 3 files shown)


mjgpy3 (commented):
Is this a good place for the following as well?
 - https://github.com/freckle/megarepo/blob/54f1e925a0d7808db4f08aee6e13fc8f3dbf7b96/backend/core/library/Freckle/Arbitrary.hs#L76
 - https://github.com/freckle/megarepo/blob/54f1e925a0d7808db4f08aee6e13fc8f3dbf7b96/backend/entities/library/Freckle/Test/Graphula.hs#L1176

mjgpy3:
Ope, sorry, you mention `selectRandom`

library/Freckle/App/Random.hs
```diff
@@ -0,0 +1,110 @@
...
+import Data.Functor ((<&>))
+import Data.STRef (newSTRef, readSTRef, writeSTRef)
+import Numeric.Natural (Natural)
+
+import qualified Data.Set as Set
+
+-- | A contiguous range of integers
+data Range i = Range
+  { inclusiveMinBound :: i
+  , size :: Natural
```
mjgpy3, line 23: Very cool model, good use of "correct by construction"

mjgpy3, line 23: Would it be a nicer model to make this "offset to end" so that empty ranges are impossible?

So `Range N 0` would mean `[N, N+0]`

chris-martin: That sounds nicer, yes.

pbrisbin:
Deferring to @mjgpy3 for review.

library/Freckle/App/Random.hs
chris-martin: I think it's good to still have an empty range possibility, but I've reworked the type definitions such that an empty range no longer has a meaningless `inclusiveMinBound` number.

mjgpy3 (approved):

PR #136, opened 2023-12-05
Use withFrozenCallStack on more exception utilities

chris-martin:
I missed a bunch of opportunities to add `withFrozenCallStack` earlier; this adds them to every definition in these modules.

Here's an example of what a snag looks like before this change.

image

After this change, the `throw` and `throwM` stack frames ought to disappear.

I've also done some eta expansions here; I'm not sure if they're necessary to make `withFrozenCallStack` work because I don't have a great intuition for how it works, but it seems more clear this way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.10.7.0...main)
 
+- Use `withFrozenCallStack` on more exception utilities to reduce noise in call stacks
+
 ## [v1.10.7.0](https://github.com/freckle/freckle-app/compare/v1.10.6.0...v1.10.7.0)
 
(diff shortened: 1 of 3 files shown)


pbrisbin (approved):
This is probably just part of the necessary change to correct the screenshot, right? I think we also want to set the Bugsnag error `message` with the `displayException` of the unwrapped exception too. Ideally, I think your screenshot wouldn't _just_ remove the extra frames, but also drop all the `AnnotatedException` noise, to look more like this:

```
user error (More than one student found with name/class combination: [...])
    library/Freckle/Api/Dal/Student.hs:99:7 throwString
```

That probably has to happen in `bugsnag` library.

pbrisbin:
The screenshot also shows me we are lacking a number of `HasCallstack` annotations, considering it shows nothing leading up to the `throwString`.

chris-martin:
Yes and yes.

PR #137, opened 2023-12-06
Improve caching errors log messages

pbrisbin:
We're seeing a lot of `OpError ValueToLarge` errors from caching set operations. This isn't necessarily a problem, but it would be good to track down and perhaps break up whatever value is being cached. Adding `error.stack` will make that possible.

As part of this, I reduced the message log-level to `WARN` as that's what they are and how we treat them. Them appearing at `ERROR` level has confused a number of engineers.

### DIFF
--- a/library/Freckle/App/Exception.hs
+++ b/library/Freckle/App/Exception.hs
@@ -1,0 +1,53 @@
+module Freckle.App.Exception
+  ( -- * Re-export of our current preferred implementation module
+
+    -- | Currently 'MonadUnliftIO'-based
+      module Freckle.App.Exception.MonadUnliftIO
+
+    -- * Helpers that are agnostic to either implementation
+  , annotatedExceptionMessage
+  , annotatedExceptionMessageFrom
+  ) where
+
+import Prelude
+
+import Control.Exception.Annotated (annotatedExceptionCallStack)
+import qualified Control.Exception.Annotated as AnnotatedException
+import Control.Monad.Logger.Aeson (Message (..), (.=))
+import Data.Aeson (object)
+import Freckle.App.Exception.MonadUnliftIO
+import GHC.Exception (prettyCallStackLines)
+
+-- | Construct a log 'Message' for any @'AnnotatedException' exception@
+--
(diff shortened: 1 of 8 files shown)


tests/Freckle/App/MemcachedSpec.hs
```diff
@@ -92,9 +93,18 @@ spec = withApp loadApp $ do
...
+      -- Object loggedMessageMeta ^? key "error" . key "stack" . _String . to T.lines
+      --   `shouldBe` Just
+      --     [ "CallStack (from HasCallStack):"
+      --     , "  throwM, called at library/Freckle/App/Memcached.hs:121:30 in freckle-app-1.10.8.0-1ebuZKUCQVI9sAWTLATGfO:Freckle.App.Memcached"
+      --     , "  cachingAs, called at library/Freckle/App/Memcached.hs:92:11 in freckle-app-1.10.8.0-1ebuZKUCQVI9sAWTLATGfO:Freckle.App.Memcached"
+      --     , "  caching, called at tests/Freckle/App/MemcachedSpec.hs:87:15 in main:Freckle.App.MemcachedSpec"
+      --     ]
+      Object loggedMessageMeta
+        ^? key "error" . key "stack" . _String . to (length . T.lines)
+        `shouldSatisfy` maybe False (> 1)
```
pbrisbin, line 110: This would be better, since it'll show the stack itself, and not just the length, on failures

```suggestion
        ^? key "error" . key "stack" . _String . to T.lines
        `shouldSatisfy` maybe False (not . null)
```

library/Freckle/App/Exception.hs
```diff
@@ -0,0 +1,53 @@
+module Freckle.App.Exception
+  ( -- * Re-export of our current preferred implementation module
+
+    -- | Currently 'MonadUnliftIO'-based
+      module Freckle.App.Exception.MonadUnliftIO
```
chris-martin, line 5: Yes, good to make that clear

chris-martin (approved):

PR #138, opened 2023-12-06
checkpoint in liftSql

chris-martin:
https://app.asana.com/0/399653121150251/1206106385053669

An easy way to help get more call stack.

`MonadIO` has to promote to `MonadUnliftIO` to catch exceptions, which I think is what @pbrisbin once suggested the constraint should be in the first place.

### DIFF
--- a/library/Freckle/App/Database.hs
+++ b/library/Freckle/App/Database.hs
@@ -65,6 +65,7 @@ import Database.PostgreSQL.Simple
 import Database.PostgreSQL.Simple.SqlQQ (sql)
 import Freckle.App.Env (Timeout (..))
 import qualified Freckle.App.Env as Env
+import Freckle.App.Exception.MonadUnliftIO
 import Freckle.App.OpenTelemetry
 import Freckle.App.Stats (HasStatsClient)
 import qualified Freckle.App.Stats as Stats
@@ -76,15 +77,15 @@ import Yesod.Core.Types (HandlerData (..), RunHandlerEnv (..))
 
 -- | A monadic context in which a SQL backend is available
 --   for running database queries
-class MonadIO m => MonadSqlBackend m where
+class MonadUnliftIO m => MonadSqlBackend m where
   getSqlBackendM :: m SqlBackend
 
-instance (HasSqlBackend r, MonadIO m) => MonadSqlBackend (ReaderT r m) where
+instance (HasSqlBackend r, MonadUnliftIO m) => MonadSqlBackend (ReaderT r m) where
   getSqlBackendM = asks getSqlBackend
 
 -- | Generalize from 'SqlPersistT' to 'MonadSqlBackend'
-liftSql :: MonadSqlBackend m => ReaderT SqlBackend m a -> m a
-liftSql (ReaderT f) = getSqlBackendM >>= f
+liftSql :: (MonadSqlBackend m, HasCallStack) => ReaderT SqlBackend m a -> m a
+liftSql (ReaderT f) = checkpointCallStack $ getSqlBackendM >>= f
 
 -- | The constraint @'MonadSqlTx' db m@ indicates that @m@ is a monadic
 --   context that can run @db@ actions, usually as a SQL transaction.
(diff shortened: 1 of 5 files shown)


library/Freckle/App/Database.hs
```diff
@@ -76,15 +78,15 @@ import Yesod.Core.Types (HandlerData (..), RunHandlerEnv (..))
 
 -- | A monadic context in which a SQL backend is available
 --   for running database queries
-class MonadIO m => MonadSqlBackend m where
+class MonadUnliftIO m => MonadSqlBackend m where
   getSqlBackendM :: m SqlBackend
```
pbrisbin, line 82: This is actually a little surprising to me. I once mentioned that `MonadTx` (or whatever it is) would make sense with a `MonadUnliftIO` constraint because it's implicitly bracket-like, but this class has no reason to implicitly want to do `MonadUnliftIO` stuff, IMO. I would've probably not even had it be `MonadIO` but only `Monad`. 

So, it would make sense to me to leave this as-is (or make it `Monad` only) and then have `(MonadUnliftIO m, MonadSqlBackend m)` on `liftSql` instead. But doesn't really matter either way, I guess.

pbrisbin (approved):
Trying to see how awesome this will be...

Does our current `runDB` call `liftSql`? Does _it_ have `HasCallStack`?

If both of those are true, I think this little change will make all `SqlError` at least have the frame of the `runDB` at minimum, which would be *epic*.

chris-martin:
> Trying to see how awesome this will be...

Not very awesome. I think. The whole process here of getting call stacks is going to be slow.

chris-martin:
There are a lot of megarepo constraints that need to promote from `MonadIO` to `MonadUnliftIO` constraints here, so not going to merge until I have a PR in for that.

pbrisbin:
> `liftSql` usage is not yet very high

Ah. Wouldn't it be higher value to do what you're doing here to `runDB` instead of `liftSql` then?

Even when `liftSql` sees adoption, it'll still be used as `runDB $ liftSql q` everywhere right? So that work would not be wasted (or even necessary)... :thinking:

chris-martin:
I believe it will be more like..

```
runSqlTx $ do
  liftSql $ ...
  liftSql $ ...
```

so both will be useful parts of the stack to have. The `liftSql` one will more likely be directly on the query, which is the frame I really want.

pbrisbin:
Gotcha. And remind me what frozen-callstack does? I thought it made is to the call itself (`liftSql` in this case) would not appear in the stack -- but we'd want it to right? I think I'm missing something there.

chris-martin:
I wasn't thinking that seeing `liftSql` in the call stack would ever be helpful, but maybe a better approach would be to be more inclusive about what's in the stacks and then later trim out what feels like noise.

pbrisbin:
Hmm, this really confuses me...

> The `liftSql` one will more likely be directly on the query, which is **the frame I really want**.
> ...
> I wasn't thinking that seeing `liftSql` in the call stack would ever be helpful

Are these not a contradiction?

pbrisbin:
Hmm, maybe we should get concrete so I'm sure I understand.

```
-- Assume everything has HasCallStack constriants

runSqlTx $ do -- 0
  fetchSomeFoo -- 1

fetchSomeFoo = do -- 3
  let x = y -- 4
  liftSql $ selectList x -- 5
```

The stack we want is,

```
liftSql:this-file:5, called at
  fetchSomeFoo:this-file:1, called at
  runSqlTx:other-files:n, called at
```

Right?

I may have this wrong, but I think using `withFrozenCallstack` on `liftSql` will result in only,

```
fetchSomeFoo:this-file:1, called at
  runSqlTx:other-files:n, called at
```

chris-martin:
I think without frozenCallStack what you'd get is

```
SqlError, raised at
  (something):in freckle-app, called at
  liftSql:this-file:5, called at
  fetchSomeFoo:this-file:1, called at
  runSqlTx:other-file:n, called at
```

but I also could be wrong.

library/Freckle/App/Database.hs
chris-martin: I would rather have it be `MonadCatch` for what it's worth.

chris-martin: Alternatively... you can always generalize further. New typeclass to abstract over annotated exception handling.

PR #139, opened 2023-12-08
Unwrap AnnotatedException to set error class and message

chris-martin:
1. `AnnotatedException` shouldn't appear in the error class
2. The annotations shouldn't appear in the exception message. If particular types of `Annotation` are desired in bugsnag, they should be added explicitly. (We presently have two types of `Annotation` that we include in bugsnag events: `CallStack` and `MetaData`.)

image


chris-martin:
Closing to do this in the `bugsnag` library instead.

PR #141, opened 2024-01-05
Fix DB metrics

joelmccracken:
- fix `active_pool_connections`
- start tracking `queries_enqueued_and_processing` where `active_pool_connections` was previously

### DIFF
--- a/library/Freckle/App/Database.hs
+++ b/library/Freckle/App/Database.hs
@@ -56,6 +56,7 @@ import Database.Persist.Postgresql
   , runSqlPoolWithExtensibleHooks
   )
 import Database.Persist.SqlBackend.SqlPoolHooks
+import Database.Persist.SqlBackend.Internal.SqlPoolHooks(SqlPoolHooks(..))
 import Database.PostgreSQL.Simple
   ( Connection
   , Only (..)
@@ -137,11 +138,41 @@ runDB
   -> m a
 runDB action = do
   pool <- asks getSqlPool
-  Stats.withGauge Stats.dbConnections $
+  gauge <- Stats.lookupGauge Stats.dbConnections
+  let
+    hooks = setAlterBackend defaultSqlPoolHooks $ wrapSqlBackend mempty
+    -- Setting the SqlPoolHooks for metrics collection:
+    -- You may be wondering if this code contains a "double-decrement" bug, because
+    -- perhaps when an exception occurs, both runAfter and runOnException are
+    -- executed. The documentation isn't terribly clear to me on this point, but
+    -- notice:
+    -- https://hackage.haskell.org/package/persistent-2.14.6.0/docs/Database-Persist-SqlBackend-Internal-SqlPoolHooks.html#t:SqlPoolHooks
+    -- especially this part regarding runOnException:
+    -- > This action is performed when an exception is received. The exception is
+    -- > provided as a convenience - it is rethrown once this cleanup function is
+    -- > complete.
+    -- Given that the exception is rethrown, then then typical runAfter handler
+    -- is very likely to not run. Looking at the code to validate:
+    -- https://hackage.haskell.org/package/persistent-2.14.6.0/docs/src/Database.Persist.Sql.Run.html#runSqlPoolWithExtensibleHooks
+    -- we can see that this is indeed how the code operates today -- the
+    -- exception is rethrown, and there aren't any other spots where other code
+    -- might catch the exception that we need to worry about, so in the case of
+    -- an exception, runAfter would not be executed.
+    -- So, this appears to be the intended interpretation.
+    hooks' = hooks
+      { runBefore = \conn mi -> do
+          Stats.incGauge gauge
+          runBefore hooks conn mi
+      , runAfter = \conn mi -> do
+          Stats.decGauge gauge
+          runAfter hooks conn mi
+      , runOnException = \conn mi e -> do
+          Stats.decGauge gauge
+          runOnException hooks conn mi e
+      }
+  Stats.withGauge Stats.dbEnqueuedAndProcessing $
     inSpan "runDB" clientSpanArguments $
-      runSqlPoolWithExtensibleHooks action pool Nothing $
-        setAlterBackend defaultSqlPoolHooks $
-          wrapSqlBackend mempty
+      runSqlPoolWithExtensibleHooks action pool Nothing hooks'
 
 runDBSimple
   :: (HasSqlPool app, MonadUnliftIO m, MonadReader app m)
--- a/library/Freckle/App/Stats.hs
+++ b/library/Freckle/App/Stats.hs
@@ -17,7 +17,11 @@ module Freckle.App.Stats
   , Gauges
   , Gauge
   , dbConnections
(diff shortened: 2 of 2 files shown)


library/Freckle/App/Stats.hs
```diff
@@ -18,6 +18,9 @@ module Freckle.App.Stats
   , Gauge
   , dbConnections
   , withGauge
+  , lookupGauge
+  , incGauge
+  , decGauge
```
pbrisbin, line 23: Sad that we now have to expose the ability to get this wrong, but I guess there's no way around it.

library/Freckle/App/Database.hs
```diff
@@ -137,11 +138,23 @@ runDB
...
+    hooks' = hooks
+      { runBefore = \conn mi -> do
+          Stats.incGauge gauge
+          runBefore hooks conn mi
+      , runAfter = \conn mi -> do
+          Stats.decGauge gauge
+          runAfter hooks conn mi
+      , runOnException = \conn mi e -> do
+          Stats.decGauge gauge
+          runOnException hooks conn mi e
```
pbrisbin, line 153: Just for future us, do you mind adding some level of detail to the commit message?

For example, we may be curious what was wrong before and how this fixes it, and I'd be interested in a link to the hooks docs. There seems to be an assumption here that the "after" hook only runs on success, requiring we also hook into "on exception". Is that documented anywhere you can link?

pbrisbin (approved):

library/Freckle/App/Database.hs
```diff
@@ -137,11 +138,23 @@ runDB
...
+      }
   Stats.withGauge Stats.dbConnections $
```
pbrisbin, line 155: Wait, you have to remove this one or we're double-counting, no?

pbrisbin (changes requested):

pbrisbin:
:facepalm: I'm sorry. I got notified of this PR being opened and thought it was a review-request. My notifications on freckle-app are a little aggressive. I now see you just opened a draft. Obviously, ignore me.

library/Freckle/App/Stats.hs
joelmccracken: Yeah I was thinking about this. There _might_ be a way to do similar to what youu suggested in a thread somewhere, but we'd need to shuffle a lot of code around/change the typeclass restrictions on certain functions. I dunno.

library/Freckle/App/Database.hs
joelmccracken: Oh yeah, I wasn't done, basically I knew that there was another metric I wanted to capture in that spot, so I kept that code there.

library/Freckle/App/Database.hs
joelmccracken: of course, was intending on doing this; at the time I wasn't totally sure that I was going to use this technique, although, actually your 

I actually added it as a comment, as I thought the area tricky enough that a comment would stand out more.

pbrisbin (approved):

library/Freckle/App/Database.hs
```diff
@@ -137,11 +138,41 @@ runDB
...
+      , runOnException = \conn mi e -> do
+          Stats.decGauge gauge
+          runOnException hooks conn mi e
+      }
+  Stats.withGauge Stats.dbEnqueuedAndProcessing $
     inSpan "runDB" clientSpanArguments $
-      runSqlPoolWithExtensibleHooks action pool Nothing $
-        setAlterBackend defaultSqlPoolHooks $
-          wrapSqlBackend mempty
+      runSqlPoolWithExtensibleHooks action pool Nothing hooks'
```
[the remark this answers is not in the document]
joelmccracken: Oh, kinda sorta. Its related, but tangental. 

I was hoping to not have to do all the mess with the hooks. Via the mechanism I propose above, I _think_ its possible, if I can work out all the details.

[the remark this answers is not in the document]
joelmccracken: What I am suggesting is the possiblity of changing all of this:
```haskell
runDB
  :: ( MonadUnliftIO m
     , MonadTracer m
     , MonadReader app m
     , HasSqlPool app
     , HasStatsClient app
     , HasCallStack
     )
  => SqlPersistT m a
  -> m a
runDB action = do
  pool <- asks getSqlPool
  gauge <- Stats.lookupGauge Stats.dbConnections
  let
    hooks = setAlterBackend defaultSqlPoolHooks $ wrapSqlBackend mempty
    -- Setting the SqlPoolHooks for metrics collection:
    -- You may be wondering if this code contains a "double-decrement" bug, because
    -- perhaps when an exception occurs, both runAfter and runOnException are
    -- executed. The documentation isn't terribly clear to me on this point, but
    -- notice:
    -- https://hackage.haskell.org/package/persistent-2.14.6.0/docs/Database-Persist-SqlBackend-Internal-SqlPoolHooks.html#t:SqlPoolHooks
    -- especially this part regarding runOnException:
    -- > This action is performed when an exception is received. The exception is
    -- > provided as a convenience - it is rethrown once this cleanup function is
    -- > complete.
    -- Given that the exception is rethrown, then then typical runAfter handler
    -- is very likely to not run. Looking at the code to validate:
    -- https://hackage.haskell.org/package/persistent-2.14.6.0/docs/src/Database.Persist.Sql.Run.html#runSqlPoolWithExtensibleHooks
    -- we can see that this is indeed how the code operates today -- the
    -- exception is rethrown, and there aren't any other spots where other code
    -- might catch the exception that we need to worry about, so in the case of
    -- an exception, runAfter would not be executed.
    -- So, this appears to be the intended interpretation.
    hooks' = hooks
      { runBefore = \conn mi -> do
          Stats.incGauge gauge
          runBefore hooks conn mi
      , runAfter = \conn mi -> do
          Stats.decGauge gauge
          runAfter hooks conn mi
      , runOnException = \conn mi e -> do
          Stats.decGauge gauge
          runOnException hooks conn mi e
      }
  Stats.withGauge Stats.dbEnqueuedAndProcessing $
    inSpan "runDB" clientSpanArguments $
      runSqlPoolWithExtensibleHooks action pool Nothing hooks'
```
to something like this:
```haskell
runDB
  :: ( MonadUnliftIO m
     , MonadTracer m
     , MonadReader app m
     , HasSqlPool app
     , HasStatsClient app
     , HasCallStack
     )
  => SqlPersistT m a
  -> m a
runDB action = do
  pool <- asks getSqlPool
  gauge <- Stats.lookupGauge Stats.dbConnections
  let
    hooks = setAlterBackend defaultSqlPoolHooks $ wrapSqlBackend mempty
    action' =
      bracket (Stats.inc theGauge) (Stats.dec theGauge) action
  Stats.withGauge Stats.dbEnqueuedAndProcessing $
    inSpan "runDB" clientSpanArguments $
      runSqlPoolWithExtensibleHooks action' pool Nothing hooks
```

not sure if that makes sense? so no need to modify `hooks` to `hooks'`, instead wrap `action`.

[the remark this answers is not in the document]
pbrisbin: :shrug: I'm too far away from this to know if that will or won't work or why. Godspeed.

PR #142, opened 2024-01-11
publish v13

joelmccracken:
I missed the finishing touches required to publish this version.

### DIFF
--- a/freckle-app.cabal
+++ b/freckle-app.cabal
@@ -1,11 +1,11 @@
 cabal-version: 1.18
 
--- This file has been generated from package.yaml by hpack version 0.35.2.
+-- This file has been generated from package.yaml by hpack version 0.35.0.
 --
 -- see: https://github.com/sol/hpack
 
 name:           freckle-app
-version:        1.12.0.0
+version:        1.12.0.1
 synopsis:       Haskell application toolkit used at Freckle
 description:    Please see README.md
 category:       Utils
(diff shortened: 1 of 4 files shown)


freckle-app.cabal
```diff
@@ -1,11 +1,11 @@
 cabal-version: 1.18
 
--- This file has been generated from package.yaml by hpack version 0.35.2.
+-- This file has been generated from package.yaml by hpack version 0.35.0.
 --
 -- see: https://github.com/sol/hpack
 
 name:           freckle-app
-version:        1.12.0.0
+version:        1.13.0.0
```
pbrisbin, line 8: This feels aggressive. Nothing about any of the code-level interfaces have changed. As far as I can tell, this is a bugfix or _maybe_ an addition. I'd go `v1.12.0.1`, myself.

pbrisbin (changes requested):

freckle-app.cabal
joelmccracken: sure, i'll do that, np. I figured adding the other metric was considered an addition, but I really _wanted_ to do this so.

pbrisbin: Even if it were and addition that'd only be `1.12.1.0`.

joelmccracken: Hmm, maybe I need to refresh on how haskell's PVP works?

pbrisbin: We have it documented here: https://illuminate.atlassian.net/wiki/spaces/EN/pages/17185952648/Maintaining+Haskell+Open+Source#Versioning

We do SemVer in the last 3 components and treat the 1st component as an overall "epoch". We like this because you can think in SemVer while still implicitly satisfying PVP, thus side-stepping a lot of flame wars about it.

pbrisbin (approved):

PR #143, opened 2024-01-26
Handle exceptions as annotated more

pbrisbin:
### [Catch query-cancelled as an AnnotatedException](https://github.com/freckle/freckle-app/pull/143/commits/412b74bea264a5c892aff2878018c4de3d4fc716)
[412b74b](https://github.com/freckle/freckle-app/pull/143/commits/412b74bea264a5c892aff2878018c4de3d4fc716)

Because of how our catch functions work, this will be behavior neutral
if the exception is not annotated (with a minor improvement in the
structure of the logged error message), but if/once it *is* annotated,
we'll include an `error.stack` value.

### [Catch exceptions as AnnotatedException in Kafka module](https://github.com/freckle/freckle-app/pull/143/commits/d066e1235890af213e9e50c711c6a0297701d291)
[d066e12](https://github.com/freckle/freckle-app/pull/143/commits/d066e1235890af213e9e50c711c6a0297701d291)

Because of how our catch functions work, this will be behavior neutral
for exceptions that not annotated (with a minor improvement in the
structure of the logged error message), but if they *are* annotated,
we'll now include an `error.stack` value.

### [Add withException to Frontrow.App.Exception](https://github.com/freckle/freckle-app/pull/143/commits/187896d37d9c0a36ff734fe63c049dacb82ab318)
[187896d](https://github.com/freckle/freckle-app/pull/143/commits/187896d37d9c0a36ff734fe63c049dacb82ab318)

### [Add logExceptionsMiddleware](https://github.com/freckle/freckle-app/pull/143/commits/9e5404c7a357bf31f409ed12ae6cd9d1428cb525)
[9e5404c](https://github.com/freckle/freckle-app/pull/143/commits/9e5404c7a357bf31f409ed12ae6cd9d1428cb525)

We currently log all handler exceptions through the `errorHandler`
member of the `Yesod` type-class. This is not ideal because it only has
access to an `InternalError Text` value, in which any exception has
already been turned into its string representation. This means we get an
ugly message like `"AnnotatedException {...}"` and aren't able to
actually construct a structured log.

We should stop doing that, and recover that logging by adding a Yesod
middleware that catches, logs, and re-throws all exceptions instead.

Then it can be handled as an actual (annotated) exception, and produce a
nice, structured log message (i.e. with `error.stack`). That's what this
middleware does. It's also what necessitated adding `withException`.

### DIFF
--- a/library/Freckle/App/Exception/MonadThrow.hs
+++ b/library/Freckle/App/Exception/MonadThrow.hs
@@ -8,6 +8,7 @@ module Freckle.App.Exception.MonadThrow
   , catches
   , try
   , tryJust
+  , withException
   , checkpoint
   , checkpointMany
   , checkpointCallStack
@@ -23,6 +24,7 @@ import Freckle.App.Exception.Types
 
 import Control.Applicative (pure)
 import Control.Exception.Annotated (checkpoint, checkpointMany)
+import Control.Monad (void)
 import Control.Monad.Catch (MonadCatch, MonadMask, MonadThrow)
 import Data.Either (Either (..))
 import Data.Function (($), (.))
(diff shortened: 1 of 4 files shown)


chris-martin (approved):
Cool!

PR #146, opened 2024-02-23
Expose 'MemcachedClient' data constructors.

StevenXL:
`fancy-api` will use `memcache` to store session data. As a result, we need to be able to require a memcache client.

By exposing the data constructors, we can extract a memcache client and continue on our merry way, or take another codebase if we do not have a client.


chris-martin (approved):
Makes sense to me. Add a note to the changelog?

PR #149, opened 2024-02-29
Add more robust commit strategy to Kafka.Consumer

pbrisbin:
Autocommit is nice and simple, but it can cause both dropped messages
and duplicate processing.

Concretely, we found that deployment of a new consumer will start and,
if the existing consumer has not (auto)committed in a bit, reprocess a
backlog of messages that it doesn't need to. This results in an increase
in lag and alerts on every deployment.

Of the various strategies available, we're implementing this one[^1]:

- When a message is processed, commit its offset asynchronously

  This is fast enough to do on every loop, but of course trades on
  reliability and may not succeed every time -- particularly if we're
  crashing or shutting down. So...

- On shutdown, commit the offset of the last-processed message (for
  every partition) synchronously

  This will be robust and cover for any failures we had from the
  asynchronous commits.

  This required we track the last message we processed on each commit as
  we go. We could've used the "commit for all messages of the last poll"
  function instead, but this would risk dropping messages if we've not
  yet processed them. We seem to only poll for 1 message at a time, so
  this risk is low, but I suspect we'll want to increase that and gain
  free performance, so it's good to be ready.

  I attempted to do the tracking with `StateT`, but it lacks instances
  like `MonadUnliftIO`, so couldn't. Instead, I went with an `IORef`. It
  is encapsulated, so we could move to something else (e.g. `TVar` or
  `Chan`) without disrupting the calling code at all, provided it only
  requires `MonadIO`.

[^1]: https://medium.com/@rramiz.rraza/kafka-programming-different-ways-to-commit-offsets-7bcd179b225a#ac4e


library/Freckle/App/Kafka/Consumer/Offsets.hs
```diff
@@ -0,0 +1,54 @@
...
+  topicPartition = topicPartitionFromMessageForCommit record
+
+commitOffsetsSync
+  :: MonadIO m
+  => LastMessageByPartition
+  -> KafkaConsumer
+  -> m (Maybe KafkaError)
+commitOffsetsSync (LastMessageByPartition ref) consumer = do
+  m <- readIORef ref
+  commitPartitionsOffsets OffsetCommit consumer $ Map.elems m
```
z0isch, line 40: Do you know the distinction between `[commitAllOffsets](https://www.stackage.org/haddock/lts-22.12/hw-kafka-client-5.3.0/Kafka-Consumer.html#v:commitAllOffsets)` and `[commitPartitionsOffsets](https://www.stackage.org/haddock/lts-22.12/hw-kafka-client-5.3.0/Kafka-Consumer.html#v:commitPartitionsOffsets)`?

If we could use `commitAllOffsets` then we wouldn't need to keep track of the offset manually but I can't figure out the distinction between them.

z0isch: Seems like it plays in tandem with `enable.auto.offset.store` from https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md

Property                                 | C/P | Range           |       Default | Importance | Description              
-----------------------------------------|-----|-----------------|--------------:|------------| --------------------------
enable.auto.offset.store | C | true, false | true | high | Automatically store offset of last message provided to application. The offset store is an in-memory store of the next offset to (auto-)commit for each partition.Type: boolean
```

pbrisbin: `commitAllOffsets` is what I meant by "commit for all messages of the last poll" in

> This required we track the last message we processed on each commit as
we go. We could've used the "commit for all messages of the last poll"
function instead, but this would risk dropping messages if we've not
yet processed them. We seem to only poll for 1 message at a time, so
this risk is low, but I suspect we'll want to increase that and gain
free performance, so it's good to be ready.

So, that'd be why I chose not to do it that way.

z0isch: Did you consider `noAutoOffsetStore` + `storeOffsetMessage` + `commitAllOffsets`? Seems like we could still get the transactionality that we want if we `storeOffsetMessage` after consuming?

pbrisbin: Ah, so it looks like we could use `enable.auto.offset.store: false` and then explicitly call [`storeOffsetMessage`](https://hackage.haskell.org/package/hw-kafka-client-5.0.0/docs/Kafka-Consumer.html#v:storeOffsetMessage) only when we've processed the message. Then `commitAlloffsets` during shutdown will only commit as far as we've stored. That sounds ideal!

z0isch: Jinx!

z0isch: These techniques still imply at least once processing, right? This should just make it way less likely.

Our consumers should really be idempotent so I think that's the correct trade-off to make, I just want to make sure I am understanding correctly.

pbrisbin: Yes, the intent is at-least-once since we cannot tolerate drops but we can tolerate duplicate processing. I think I've achieved it?

z0isch (approved):
❤️

PR #151, opened 2024-03-06
Batch Kafka consumer

z0isch:
Add ability to handle batches for our kafka consumers

### DIFF
--- a/library/Freckle/App/Kafka/Consumer.hs
+++ b/library/Freckle/App/Kafka/Consumer.hs
@@ -164,31 +164,39 @@ runConsumer
      , HasCallStack
      )
   => Timeout
+  -> Int
   -> (a -> m ())
   -> m ()
-runConsumer pollTimeout onMessage =
+runConsumer pollTimeout batchSize onMessage =
   withTraceIdContext $ immortalCreate onFinish $ do
     consumer <- view kafkaConsumerL
 
     flip catches handlers $ inSpan "kafka.consumer" consumerSpanArguments $ do
-      mRecord <- fromKafkaError =<< pollMessage consumer kTimeout
+      mRecords <-
+        traverse fromKafkaError
+          =<< pollMessageBatch consumer kTimeout kBatchSize
 
-      for_ mRecord $ \r -> do
-        for_ (crValue r) $ \bs -> do
-          a <-
-            inSpan "kafka.consumer.message.decode" defaultSpanArguments $
-              either (throwM . KafkaMessageDecodeError bs) pure $
-                eitherDecodeStrict bs
-          inSpan "kafka.consumer.message.handle" defaultSpanArguments $ onMessage a
+      for_ mRecords $ \mRecord -> do
+        for_ mRecord $ \r -> do
+          for_ (crValue r) $ \bs -> do
+            a <-
+              inSpan "kafka.consumer.message.decode" defaultSpanArguments $
+                either (throwM . KafkaMessageDecodeError bs) pure $
+                  eitherDecodeStrict bs
+            inSpan "kafka.consumer.message.handle" defaultSpanArguments $ onMessage a
+            logExMay
+              "Unable to store offset"
+              (silenceNoOffsetError <$> storeOffsetMessage consumer r)
 
-        inSpan "kafka.consumer.message.commit" defaultSpanArguments $ do
-          -- Store the offset of this record, then do a best-offort commit to
-          -- the broker. If this fails and we crash or shutdown, the commit in
-          -- the onFinish handler will pick it up.
-          logExMay "Unable to store offset" $ storeOffsetMessage consumer r
-          void $ commitAllOffsets OffsetCommitAsync consumer
+      inSpan "kafka.consumer.message.commit" defaultSpanArguments $
+        -- Do a best-effort commit to the broker. If this fails and
+        -- we crash or shutdown, the commit in the onFinish handler
+        -- will pick it up.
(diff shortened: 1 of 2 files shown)


library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -164,31 +164,39 @@ runConsumer
...
+            logExMay
+              "Unable to store offset"
+              (silenceNoOffsetError <$> storeOffsetMessage consumer r)
+
+      inSpan "kafka.consumer.message.commit" defaultSpanArguments $
+        -- Do a best-effort commit to the broker. If this fails and
+        -- we crash or shutdown, the commit in the onFinish handler
+        -- will pick it up.
+        void $
+          commitAllOffsets OffsetCommitAsync consumer
```
pbrisbin, line 196: I believe the `_NO_OFFSETS` error comes from here (when you commit with nothing stored), so you'd want to move your silence function, right?

library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -164,31 +164,39 @@ runConsumer
      , HasCallStack
      )
   => Timeout
+  -> Int
   -> (a -> m ())
```
z0isch, line 168: Is this the right shape for the handler? Should it be `[a] -> m ()` instead? 

Seems like it'd more flexible for the client to do "smart" things if we hand them the entire batch. I'm thinking things like doing concurrent things or grouping stuff together.

pbrisbin (approved):
So this re-introduces some duplicate processing. If we're deploying a second consumer can come up and would reprocess any messages the existing consumers have stored but not yet committed, which is up to `batchSize` messages.

Seems like an OK middle-ground with out it was before. However, I still feel like we could've arrived at the same rough behavior in this regard if we had simply reduced the auto-commit setting to do so more frequently, no?

library/Freckle/App/Kafka/Consumer.hs
z0isch: I'm seeing the `Unable to commit offsets` message which I think comes from the `logExMay "Unable to store offset"` when we `storeOffsetMessage`

pbrisbin:
If we wanted to keep the reduction in duplicate processing, I think still doing a store+commit on every message, but now with batching could also help things... since you're polling less frequently it would take pressure off the broken even though you're committing more frequently. WDYT?

library/Freckle/App/Kafka/Consumer.hs
pbrisbin: Sorry, I'm confused... you're saying you see `"Unable to commit offsets"` but you claim it comes from where we log `"Unable to store offset"`? That doesn't seem right, those are different messages.

Log message have source location though, so we shouldn't have to guess.

library/Freckle/App/Kafka/Consumer.hs
pbrisbin: Good question. I defer to you.

Practically speaking, I'd put my money on consumers _not_ doing smart things, but it's also trivial for them to do their own `traverse_`.

z0isch: I think `[a] -> m ()` also implies that we can't store or commit offsets on each event, so I guess for the first attempt at this I can go with the `a -> m ()` shape and see what happens.

z0isch:
> If we wanted to keep the reduction in duplicate processing, I think still doing a store+commit on every message, but now with batching, could also help things... since you're polling less frequently it would take pressure off the broker even though you're committing more frequently. WDYT?

Probably worth a try to see what happens; however a client could still get this behavior by having `batchSize=1` right?

library/Freckle/App/Kafka/Consumer.hs
z0isch: Oh the looping is from `immortal`- we call `onFinish` after every message

z0isch: Wait- that means we have a synchronous offset commit on every message. Is that what we really want?

z0isch: Opened up https://github.com/freckle/freckle-app/pull/152 to explore fixing this

z0isch:
Closing this as the performance is back to normal with doing auto-commits for the answers consumer.

It seems like there is just a trade off on size of the batch of offset commits vs. reprocessing events. For our current consumer re-processing is fine due to the idempotency of the consumer.

It also seems like there is just going to be a performance cost to pay when deploying new consumers that is un-related to reprocessing events. The performance hit seems likely due to the kafka rebalance when adding/removing consumers to a consumer group and not due to reprocessing a handful (100s of events).

pbrisbin:
> It also seems like there is just going to be a performance cost to pay when deploying new consumers that is un-related to reprocessing events

Did slowing the rotation from double-then-halve to plus-one-minus-one solve it?

z0isch:
> > It also seems like there is just going to be a performance cost to pay when deploying new consumers that is un-related to reprocessing events
> 
> Did slowing the rotation from double-then-halve to plus-one-minus-one solve it?

Nope, it just made it have 3 blips instead of one. I think this makes sense because there was probably 3 re-balances
image

AFAICT it seems like the only way to avoid these performance issues is to not rebalance which I think implies doing some sort of static partition assignment for the consumers. This would require a lot of complexity in coordination between the consumers though, so I think we'd need to weight that against how important it is to not have these spikes.

PR #152, opened 2024-03-06
Explicitly loop in runConsumer

z0isch:
We were relying on `immortal` to handle the looping for `runConsumer` which calls the `onFinish` after every loop. This makes our strategy of calling a synchronous offset commit happen on every loop instead of just on error/shutdown of the consumer.

This changes to explicitly loop in the consumer so that `onFinish` only gets called on shutdown/crashes.

### DIFF
--- a/library/Freckle/App/Kafka/Consumer.hs
+++ b/library/Freckle/App/Kafka/Consumer.hs
@@ -12,6 +12,7 @@ import Freckle.App.Prelude
 
 import Blammo.Logging
 import Control.Lens (Lens', view)
+import Control.Monad (forever)
 import Data.Aeson
 import Data.ByteString (ByteString)
 import qualified Data.List.NonEmpty as NE
@@ -167,9 +168,7 @@ runConsumer
   -> (a -> m ())
   -> m ()
 runConsumer pollTimeout onMessage =
-  withTraceIdContext $ immortalCreate onFinish messageLoop
- where
-  messageLoop = do
+  withTraceIdContext $ immortalCreate onFinish $ forever $ do
     consumer <- view kafkaConsumerL
 
     flip catches handlers $ inSpan "kafka.consumer" consumerSpanArguments $ do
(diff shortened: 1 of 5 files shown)


library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -167,7 +167,9 @@ runConsumer
   -> (a -> m ())
   -> m ()
 runConsumer pollTimeout onMessage =
-  withTraceIdContext $ immortalCreate onFinish $ do
+  withTraceIdContext $ immortalCreate onFinish messageLoop
```
pbrisbin, line 170: Why not just?

```suggestion
  withTraceIdContext $ immortalCreate onFinish $ forever $ do
```

library/Freckle/App/Kafka/Consumer.hs
```diff
@@ -203,7 +206,9 @@ runConsumer pollTimeout onMessage =
 
   onFinish finishResult = do
     consumer <- view kafkaConsumerL
-    logExMay "Unable to commit offsets" $ commitAllOffsets OffsetCommit consumer
+    logExMay
+      "Unable to commit offsets"
+      (silenceNoOffsetError <$> commitAllOffsets OffsetCommit consumer)
```
pbrisbin, line 211: Told ya?

pbrisbin (approved):
Ugh, great catch. I always misinterpret `Immortal` here. It does nothing to ensure you're process actually _is_ a "forever" process itself.

Should our `immortalCreate` just always do that for you? Or is is that too far afield from the upstream behavior to "fix" for everyone?

library/Freckle/App/Kafka/Consumer.hs
z0isch: I always forget about `forever` for some reason...

library/Freckle/App/Kafka/Consumer.hs
z0isch: mhmm.... It's pretty sweet that it pointed us to this issue

z0isch:
> Should our immortalCreate just always do that for you? Or is is that too far afield from the upstream behavior to "fix" for everyone? 

I think you mean "fixm" 😉 

I really don't know- how bad would it be to `immortalCreate onFinish $ forever $ ...` if `immortalCreate` also did a `forever`?

PR #153, opened 2024-03-07
Introduce MonadHttp, add Freckle.App.Test.Http

pbrisbin:
This commit re-writes `Freckle.App.Http` to use a `MonadHttp` type class
instead of being in `MonadIO`. This type class will allow us to
instantiate differently in different contexts, most usefully with
stubbed responses in test.

### `MonadHttp`

The `MonadHttp IO` instance is the same behavior as the function before,
this means users can maintain behavior-neutrality across this upgrade by
either giving their own monad stack the instance,

```hs
instance MonadHttp (AppT m) where
  httpLbs = liftIO . httpLbs

-- Or, in a Yesod context
instance MonadHttp (HandlerFor App) where
  httpLbs = liftIO . httpLbs
```

Or by changing all invocations to use `IO`:

```diff
- resp <- httpLbs ...
+ resp <- liftIO $ httpLbs ...
```

### Effects on Other Modules

`Freckle.App.Http.Paginate` was relaxed to `Monad m`, which could've
always been the case, since it takes the "request runner" as argument.

`Freckle.App.Http.Retry` still requires `MonadIO` since it's
(annoyingly) exception based, owing to the fact that we can't assume a
non-raising `checkResponse` member on the `Request`s we're running. NOTE
that the `IO` instance for `MonadHttp` uses `rateLimited`, since that's
how the previous `httpLbs` worked.

### `HttpStub`s

The principle reason for doing this was to introduce a way to stub this
`httpLbs` to return with pre-defined responses, which is what
`Freckle.App.Test.Http` is all about.

This is accomplished by holding a list of `HttpStub` somewhere, which
defines how to respond to requests that match. The simplest way to do so
is to use the `IsString` instance:

```hs
stubs :: [HttpStub]
stubs =
  [ "https://example.com"
  ]
```

When run in a context where these stubs are involved, this will respond
200 with an empty body for any `GET` requests made to this domain. The
response can be changed through lenses:

```hs
stubs :: [HttpStub]
stubs =
  [ "https://example.com"
      & statusL .~ status400
      & bodyL .~ "Let's test a Bad Request"
  ]
```

The string is passed to `parseRequest_`, so anything valid there is
valid here:

```hs
data MyItem = MyItem
  { -- ...
  }
  deriving stock Generic
  deriving anyclass ToJSON

stubs :: [HttpStub]
stubs =
  [ "POST https://example.com/items"
      & json [MyItem]
      -- ^ Now matches requests with JSON in the Accept Header only
      --   Responds with Content-Type JSON
      --   Responds with a body of the JSON-encoded items
  ]
```

For additional usage details see the Haddock comments throughout.

Once we have the `stubs` we can set up a `MonadHttp` context that uses
it:

```hs
data TestApp = TestApp
  { appHttpStubs :: [HttpStubs]
  }

-- Assume TestAppT is a ReaderT TestAppt
instance MonadHttp (TestAppT m a) where
  httpLbs req = do
    stubs <- asks appHttpStubs
    pure $ httpStubbed stubs req
```

Additionally, there are also tools for `DerivingVia` or running things
in a concrete `HttpStubsT` stack.

### Handling Un-stubbed Requests

When no stubs match a give request we call `error` -- this seems
uncouth, but is actually the best possible behavior for the intended
use-case in (e.g.) HSpec:

One other reasonable behavior would be to respond 404 to any un-matched
requests. This can be accomplished by adding a "match anything" stub at
the end:

```hs
stubs :: [HttpStub]
stubs =
  [ -- ...
  , -- ...
  , httpStub "Anything" MatchAnything
      & statusL .~ status404
      & bodyL .~ "Not found"
  ]
```


pbrisbin:
I used `NoFieldSelectors` and `OverloadedRecordDot` in my new modules because we're moving that way in our conventions and I'm finding it more and more natural (and preferred). However, that breaks build for GHC < 9.0. In order to ship this code as-is, we would have to drop that support.

I can say with confidence that in most of our OSS we would not want to drop that support, we should aim for our OSS to be broadly useful. I can say with confidence that we don't need that support in our Apps, our apps always run latest and we are far enough along and (as a team) in favor of these extensions. `freckle-app` sits in-between, so I don't know how we should handle this question here.

I'm going to request some early review, but will keep this Draft until we answer that question.

chris-martin:
I'm of the opinion that maintaining <9 on anything is a waste of our effort, but freckle-app especially.

library/Freckle/App/Test/Http.hs
```diff
@@ -0,0 +1,332 @@
...
+      <> show req
+      <> "\n"
+      <> concatMap (uncurry unmatchedMessage) unmatched
+
+  unmatchedMessage stub err = "\n== " <> stub.label <> " ==\n" <> err
+
+-- | Fields that can be defined for a response
+data HttpStubResponse = HttpStubResponse
+  { status :: Status
+  , headers :: ResponseHeaders
```
chris-martin, line 107: What do you think about abstracting this slightly further as `headers :: Map (CI ByteString) ByteString`?

library/Freckle/App/Test/Http.hs
```diff
@@ -0,0 +1,332 @@
...
+
+-- | Modify the stub to match JSON requests and respond with the given value
+json :: ToJSON a => a -> HttpStub -> HttpStub
+json a stub =
+  stub
+    & matchL <>~ MatchHeader (hAccept, "application/json")
+    & headersL <>~ [(hContentType, "application/json")]
+    & bodyL .~ encode a
+
+-- | Load stubs from the filesystem
```
chris-martin, line 182: Very cool

library/Freckle/App/Test/Http.hs
pbrisbin: Well, I think it's generally best to match the type of the same field in `Response`, such to ensure we can represent any possible `Response` in the wild.

I had to think about this case in particular, and it does happen to matter. Not only does HTTP allow multiple values at the same header name, but there are actual semantics for it too: https://stackoverflow.com/a/4371395.

chris-martin:
One further thing I can imagine wanting is the ability to define stubs with side effects. I don't want to get into talking about full-fledged mocking of stateful services, but just in terms of laziness of the test data - I wish `loadHttpStubsDirectory` didn't load all the files into memory upfront, and when I want to test with fake data I wish it could be generated on demand. If we're going to take some of the arbitrary data generation techniques we currently use with Graphula and translate that from Postgres to HTTP, then we'll need the ability to write stubs backed by some mutable store.

library/Freckle/App/Test/Http.hs
chris-martin: The semantics for it are exactly why it seems reasonable to say we should be able to normalize responses that do have multiple fields of the same name into a standard form that doesn't.

pbrisbin: I think you're implying that normalizing the representation is better than just remaining faithful to the representation originally stubbed. Is that right? If so, why?

pbrisbin:
> I wish loadHttpStubsDirectory didn't load all the files into memory upfront, and when I want to test with fake data I wish it could be generated on demand

It seems there are 3 distinct features...

1. Stateful stubs

   So this basically means Mocks. Which is something I'll want, but not what this intends to be. Being able to expect an HTTP call and fail if it doesn't happen, being able to call the same path multiple times (and get 429, 429, 200), etc. These would require a different (stateful) mechanism. We'll see, but it's intentional this module is not that.

2. Effectful stubs

   Being able to stub a request to "/foo.json" to then read `foo.json` from the filesystem when requested is definitely possible and would not complicate things much.

3. Stubs defined from effects like arbitrary

   This is already possible, I've converted a number of megarepo specs that rely on it already. You just run the affect first to get the `[HttpStub]`.

   ```hs
   stubs <- runIO setupStubs

   withApp loadApp $ modifyApp (httpStubsL ?~ stubs) $ do
    -- ...

   setupStubs :: [HttpStub]
   setupStubs = do
     questions <- generate $ vectorOf 10 arbitrary
     pure [".../questions.json" & json @[Question] questions
   ```

   In fact, `loadStubsDirectory` is just applying this pattern.

So, I guess, given that (1) is an anti-design goal, (3) is already doable, is (2) worth exploring?

library/Freckle/App/Test/Http.hs
chris-martin: Haven't really thought through why, it's just it seems like the purpose of this type already is to be a slightly higher-level representation of a response, as it doesn't represent the http version, exotic bits like trailer fields, etc. so thought I'd throw it out there.

chris-martin:
> Which is something I'll want, but not what this intends to be.

Fair enough. The `MonadHttp` class is the widest generalization and one can swap in something different if `HttpStub` isn't the right thing for a case.

library/Freckle/App/Test/Http.hs
pbrisbin: :shrug:

chris-martin (approved):

PR #155, opened 2024-03-08
Various breaking changes

pbrisbin:
Since `MonadHttp` introduces a breaking change (one that is actually kind of annoying and
not just happening for pedantic reasons), and required dropping some older resolvers from
our CI testing, I figured we should batch some waiting stuff to go along with it:

- Remove deprecated function
- Derive MonadUnliftIO
- Remove Dotenv.Compat module
- Remove HTTP.Link.Compat module
- Remove Freckle.App.Aeson compat module
- Remove Env.kept
- Assume memcache-0.3.0
- Assume template-haskell-2.18.0
- Update stack-nightly extra-deps
- Silence new warning in ghc-9.8
- Compile with aeson-2.2
- Re-organize stack.yamls
- Support http-client-0.7.16
- Document changes


chris-martin:
Not a fan of the [give the user the option to upgrade in smaller steps](https://www.haskellforall.com/2019/05/release-early-and-often.html) philosophy?

pbrisbin:
I'm not sure that's the choice here. Once we dropped support for GHC < 9.0, and stopped CI'ing with lts-19.33, all of the CPP I removed became dead code. 1.15 would cause that whether I include all this or not. What would doing these changes in smaller increments (what I think you're suggesting) accomplish?

chris-martin:
Oh I see, I missed that dropping older GHCs was the impetus. That makes more sense.

library/Freckle/App/Aeson.hs
```diff
@@ -1,26 +0,0 @@
-{-# LANGUAGE CPP #-}
-
--- | Compatibility module for "Data.Aeson" 1.x vs 2.0
```
chris-martin, line 3: :partying_face:

chris-martin (approved):

PR #156, opened 2024-03-11
Freckle.App.Http.Cache

pbrisbin:
This adds a function `Request -> Response body` that caches based on
all the same rules as Browsers or CDNs. For example, key by URL + `Vary`, set
TTL based on `Cache-Control[max-age]` or `Expires`, respect `no-store` and
`private`, etc.

The function takes a relatively complex settings type, but one is defined for you
that caches into `Memcached`, serializing the response as CBOR. This code
somewhat duplicates what we have in `Freckle.App.Memcached`, because
the idea of setting TTL dynamically based on the response was not easily
supported there.

In addition to this implementation, I also built one using `StateT` and an
in-memory map. This was necessary in order to (purely) test the caching logic itself, but
could also be used "for real" in non-production environments, provided you
can construct things to satisfy `(MonadState s m, HasCache s)`.

The function can wrap `MonadHttp(httpLbs)` or even be used in the
implementation there-of, in order have all `MonadHttp` functions throughout
your app get this caching all the time without having to do anything specific at
the call-site.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206639636897610


chris-martin:
Hey, somewhat related thought - wdyt about adding an environment variable corresponding to the `max_item_size` setting in elasticache (currently set at the recommended default of 1,048,576 bytes) so that, before we attempt to write a memcache object, if it's too large we can skip the write that's going to fail and send a warning to bugsnag instead?

chris-martin:
I could imagine this being popular as a standalone package.

pbrisbin:
> if it's too large we can skip the write that's going to fail and send a warning to bugsnag instead?

Would it be more robust / useful to just provide a hook to send any warnings to Bugsnag (or anything else) and then let those failures report there along with any others?

> I could imagine this being popular as a standalone package.

100%, I think all of `Freckle.Http.*` should be spun off.

library/Freckle/App/Http/Cache.hs
```diff
@@ -0,0 +1,304 @@
...
+--      , 'MonadHttp' m
+--      , 'MonadLogger' m
+--      , 'MonadTracer' m
+--      , 'MonadReader' env m
+--      , 'HasMemcachedClient' env
+--      )
+--   => FooId -> m [Foo]
+-- getFooById fooId = do
+--   let req = parseRequest_ $ \"https:\/\/api.com\/foos\/\" <> fooId
+--   'httpCached' ('memcachedHttpCacheSettings' 'fiveMinuteTTL') 'httpJson' req
```
pbrisbin, line 114: I don't _actually_ know if this works yet. It would need `Serialise (Either DecodeError a)`, but if that instance could exist, then passing `httpJson` as the `Request -> m (Response body)` like this would totally work.

library/Freckle/App/Http/Cache.hs
```diff
@@ -0,0 +1,304 @@
...
+          Left err -> doHttpSet now key <* settings.logWarn err
+          Right cresp | isCachedResponseStale cresp now -> do
+            case lookupHeader hETag cresp.response of
+              Nothing -> evictAndRecache now key =<< doHttp req
+              Just etag -> do
+                resp <- doHttp $ addRequestHeader hIfNoneMatch etag req
+                case statusCode (getResponseStatus resp) of
+                  304 -> pure cresp.response
+                  _ -> evictAndRecache now key resp
+          Right cresp -> pure cresp.response
```
pbrisbin, line 143: I tried writing this so many different ways, and none of them were particularly pleasing beyond this straight-forward stair-`case`-ing. I think there's really no hope of making this code super clear. Instead, I recommend reading the tests to understand the behaviors.

library/Freckle/App/Http/Cache/Memcached.hs
```diff
@@ -0,0 +1,136 @@
...
+     , MonadTracer m
+     , MonadReader env m
+     , HasMemcachedClient env
+     , Serialise body
+     )
+  => CacheTTL
+  -- ^ Default TTL, used when @max-age@ is not present
+  -> HttpCacheSettings m body Value
+memcachedHttpCacheSettings defaultTTL = HttpCacheSettings
+  { shared = False
```
pbrisbin, line 43: Should be `True`. This field used to be called `private` so I keep getting the meaning flipped in my head. Only a local Browser cache could really be considered un-shared.

library/Freckle/App/Memcached/Client.hs
```diff
@@ -95,6 +96,22 @@ set k v expiration = traced $ with $ \case
               ]
         }
 
+-- | Delete a key
+delete
+  :: (MonadUnliftIO m, MonadTracer m, MonadReader env m, HasMemcachedClient env)
+  => CacheKey
+  -> m ()
+delete k = traced $ with $ \case
+  MemcachedClient mc -> void $ liftIO $ Memcache.delete mc (fromCacheKey k) 0
```
pbrisbin, line 105: **Big TODO**: this is passing `0` for the version argument (you have to say which version you want to delete), which doesn't work. I think there is no option other than to do a `get`-`delete` here, but at least we could hide it within this function.

library/Freckle/App/Http/Cache/Memcached.hs
```diff
@@ -0,0 +1,144 @@
...
+
+memcachedHttpCodec :: Serialise body => HttpCacheCodec body Value
+memcachedHttpCodec =
+  HttpCacheCodec
+    { serialise = BSL.toStrict . serialise . fromResponse
+    , deserialise = \req ->
+        bimap show (toResponse req)
+          . deserialiseOrFail
+          . BSL.fromStrict
+    }
```
chris-martin, line 61: There is a subtle concern I want to think about carefully here.

So (nothing in this paragraph you don't already know but), a web service like assets.freckle is an API for which we have to think about backwards compatibility. We do this by not removing things unless we know they're not in use, by taking advantage of the fact that adding a new endpoint is a non-breaking changes, and by the fact that adding new fields to JSON objects is a non-breaking change.

The thing that I don't see talked about as much: A cache layer on top of such a service is *another* API that also has to preserve backwards compatibility. It's not as big a deal because things only stay in the cache for five minutes, but it still means that a breaking change to a `Serialise` instance could easily produce a five-minute outage when we deploy a new version of fancy-api that expects a different serialization format from what's in memcache. Or worse, maybe we get unlucky enough that deserialization actually succeeds and ends up producing erroneous results.

The nice property of JSON that adding new fields is a non-breaking change is - I doubt? - something that holds for CBOR. So I think a system for versioning this API - to make sure that the cache keys change when the value format changes - is really worth considering.

This is a big part of why I'm coming around to the idea of caching exactly the server's response rather than transforming it first, because whatever that transformation is, it's something that also needs to be versioned. Which gets really really hard when you're dealing with nested records of types with `Serialise` instances, because, just like with JSON, a change to a type in one place can easily end up having an action at a distance on some API that you didn't know was versioned.

I'm not sure I've thought this through all the way, but my thought is that `Serialise` is a footgun here that we should avoid in favor of something value-level codec type which represents a `Serialise` instance plus a version identifier, like

```haskell
data Codec a = Codec
  { encode :: a -> Encoding
  , decode :: Decoder s a
  , version :: Tree Text
  }
```

```haskell
memcachedHttpCodec :: Codec body -> HttpCacheCodec body Value
```

... `version` being a `Tree` where the root node is the version of your codec, and the children are version identifiers from the codecs that you use. Probably a hash of that tree then ends up in the cache keys.

I was going for something similar with [cacheKeyContribution](https://github.com/freckle/content-client/blob/a840a4f8b1a531f7f9f6e6fd820f468f89d91984/content-client-core/library/Freckle/Content/Client/Core/Abstract/Exchange.hs#L65) in content-client - The idea being that when you change types, you should increment that value to result in moving to new cache keys - but I don't think what I did there was robust enough to solve the problem, because I'm still using a lot of generic-derived `Serialise` instances and I don't imagine that anyone modifying types in the library will remember to go looking for versions to bump according to the resulting cascading changes to the CBOR API that we never see or think about.

library/Freckle/App/Http/Cache/Memcached.hs
```diff
@@ -0,0 +1,144 @@
...
+  , cinserted :: UTCTime
+  , cttl :: CacheTTL
+  }
+  deriving stock (Generic)
+  deriving anyclass (Serialise)
+
+{- FOURMOLU_DISABLE -}
+-- Fourmolu has trouble with this bit of CPP
+
+toResponse :: Request -> CResponse body -> CachedResponse body
```
chris-martin, line 93: I am having trouble remembering the distinction between `CResponse` and `CachedResponse`, I don't know if I need different names or just comments or what.

library/Freckle/App/Http/Cache/Memcached.hs
```diff
@@ -0,0 +1,144 @@
...
+    , set = \k v -> try $ Memcached.set k v 0
+    , evict = try . Memcached.delete
+    }
+
+data CResponse body = CResponse
+  { cresponseStatus :: Status
+  , cresponseVersion :: HttpVersion
+  , cresponseHeaders :: ResponseHeaders
+  , cresponseBody :: body
+  , cresponseCookieJar :: CookieJar
```
chris-martin, line 82: Why are cookies here at all? That does not seem like something you'd care to cache. I am only able to imagine why this could be bad and not why it could be good.

pbrisbin:
> I have no idea what happens when you try to store an oversized memcache value. Would we get an error report now? Would it fail silently? I have no idea.

`Freckle.App.Memcached` catches all exceptions and logs them using `MonadLogger`. That doesn't matter here because none of this module is using that facility (since it doesn't support dynamic TTL).

`Freckle.App.Http.Cache` expects all cache operations to return `Either SomeException`, which it then logs using `settings.logWarn . displayException`. `Freckle.App.Http.Cache.Memcached` implements `logWarn` to be that of `MonadLogger`.

Setting a too-large value is an exception, which would hit `settings.logWarn`. So it gets logged and otherwise ignored. This is in keeping with "caching is best effort". Behavior we of course don't want in sessions storage ;)

> I'm not sure exactly what you mean or how it's different from what I said.

My understanding of what you said would be to explicitly check for the size situation and separately notify Bugsnag. I'm suggesting we don't do anything special and instead allow one to make a `settings.logWarn` that is "notify bugsnag", if they want that. This would result in _all_ exceptions (including value-too-big) to notify Bugsnag. Just changing it from `logWarn :: String -> m ()` to `logWarn :: SomeException -> m ()` should be enough.

library/Freckle/App/Http/Cache/Memcached.hs
pbrisbin: Barring any other trade-offs, I was just trying to cache as much as I could to keep a cached response and a real response as identical as I can. I feel like that's still the right thing to do.

Why would it be bad?

library/Freckle/App/Http/Cache/Memcached.hs
pbrisbin: It is annoying. `CachedResponse` is the value that goes into cache (any cache). And `CResponse` is just something we have to convert that to in order to give it a `Serialise` instance.

Things would be a lot easier if I could just give `Response body` an orphan instance, and then derive `Serialise` for `CachedResponse`. But the trouble is `responseOriginalRequest` -- get `Serialise` on `Request` looks impossible (it has a few functional fields), so we simply can't serialise an actual `Response`. That's why `CResponse` omits it and then is given in back again when deserialising.

pbrisbin: Would `s/CResponse/SerialisableResponse/` work?

library/Freckle/App/Http/Cache/Memcached.hs
chris-martin: It seems like what would be most bad is if we ever actually used those cookies, creating some complex consequences of trying to cache requests within stateful sessions? I don't know. Vague unknown-unknown concern.

Only concrete concern is - why slightly inflate the cache size with information that is redundant to the response headers?

library/Freckle/App/Http/Cache.hs
```diff
@@ -0,0 +1,304 @@
...
+ where
+  handleCachableRequest key = do
+    now <- settings.getCurrentTime
+    result <- fromEx Nothing $ settings.cache.get key
+
+    case result of
+      Nothing -> doHttpSet now key
+      Just val ->
+        case settings.codec.deserialise req val of
+          Left err -> doHttpSet now key <* settings.logWarn err
```
chris-martin, line 134: > Caching is best-effort with failures (such as deserialisation) log-ignored

Yes, I see.

library/Freckle/App/Http/Cache/Memcached.hs
pbrisbin: > I guess I didn't mean raw bytes on the wire, but rather that it's easier to get right if there is a single mapping from http-client's response type to our cached format rather than bespoke formats for each route that experience a lot of minor shifts all the time.

Are you just suggesting we use `Response ByteString` here instead of `SomeConstraint body => Reponse body`?

I'm fine with that, it's simpler and is how it worked in an earlier version. I thought you described caching a more compact format, post-JSON processing as somewhat of a goal, so I was reaching for it here.

> It is, that's the point

Gotcha. Yeah, so what I'm saying is you could take on all that complexity to implement such a thing, or you could just lean on the "de-serialization failures behave like a cache miss" that we have today and the behavior is the same. Better even, since the latter will still hit when it can. The former requires developers remember to take an extra action, which they'll forget, and it will _always_ generate misses, whether there actually is a format change that impacted your cached values or not.

library/Freckle/App/Http/Cache/Memcached.hs
pbrisbin: > that is redundant to the response headers

Yeah, that's my mistake. For some reason, I was picturing something like `Set-Cookie` as information that is in the Cookie Jar, and so retaining it would be necessary to be a faithful representation of the response. But you're right, obviously that's all in the headers and the Jar is a client state thing added by http-client that has nothing to do with information coming from the server in this response.

I will remove it.

library/Freckle/App/Http/Cache/Memcached.hs
chris-martin: > I thought you described caching a more compact format, post-JSON processing as somewhat of a goal

I did, but it seems less desirable as a wild west of myriad datatypes with unversioned binary formats.

chris-martin: > I'd like to understand more about what kind of breaking changes occur in Serialise and how likely valid-but-different really is though, before I jump off that cliff.

Any time you change the order of fields of the same type in a record you're going to have this hazard, for one thing.

```haskell
data Person_v1 = Person_v1 { name :: Text, password :: Text }
  deriving stock (Generic, Show) deriving anyclass Serialise
data Person_v2 = Person_v2 { password :: Text, name :: Text }
  deriving stock (Generic, Show) deriving anyclass Serialise
```

```haskell
ghci> cachedValue = serialise Person_v1{ name="Chris", password="xyz" }
ghci> deserialise @Person_v2 cachedValue
Person_v2 {password = "Chris", name = "xyz"}
```

pbrisbin: OK, yeah that's 100% a deal-breaker for anything other than `Response ByteString` as our target. Done.

chris-martin:
> My understanding of what you said would be to explicitly check for the size situation and separately notify Bugsnag.

Yeah the "separately notify bugsnag" was a distraction, the point was "explicitly check for the size situation" - Trying to ask whether there's any advantage to the client knowing what the max size is so that it can fail fast without hitting memcache with an operation that is certain to fail. Probably an unimportant optimization though.

library/Freckle/App/Http/Cache/Memcached.hs
pbrisbin: FWIW, I was talking about upgrading `serialise` the library not actually run-of-the-mill changes to your own datatypes. That is why I didn't realize that just backing off to storing `Response ByteString` was a fix, since library updates still impact that.

Changing your own types would be waaaaaay more common situation, which (now that I see what you mean) is not something we could ignore.

pbrisbin:
> the point was "explicitly check for the size situation" - Trying to ask whether there's any advantage to the client knowing what the max size is so that it can fail fast without hitting memcache with an operation that is knowable beforehand to be doomed to failure

Gotcha.

> Probably an unimportant optimization though

I think so :shrug: 

How does one actually answer? I think we'd need to know:

- Roughly how long does a successful memcached set take
- Roughly how long does an unsuccessful memcached set take

It would only make sense, IMO, if the latter is much slower. Otherwise, you'd be optimizing away time that clients were always willing to pay anyway. Otherwise it's not worth any complexity on our end, IMO.

I doubt that the time spent for the error case is substantially slower than success (I'd guess the opposite actually), but I don't know.

library/Freckle/App/Http/Cache/Memcached.hs
chris-martin: I think we talked about this before but I don't recall exactly - Any transport compression on that ByteString has been undone at this phase, right? Maybe we should apply our own compression, store the responses gzipped or something? Could make a big difference on some of the larger JSON responses.

chris-martin: fwiw `math/en/questions.json` is 63 meg raw, 5 meg gzipped.

pbrisbin: I don't think that's handled at all at this level. However you would make your request and interpret the returned `ByteString` today is how it would work with this caching in place.

For example, if you made a request with `Accept: gzip` with `Network.HTTP.Client`, I'm relatively certain you need to decompression that response body you get back. That would be unchanged by putting this cache in between.

I agree it could be valuable to automatically add that `Accept` and decompress (based on the returned `Content-Type`) so that our cached values are smaller. Maybe we explore adding that later / next?

chris-martin: > if you made a request with Accept: gzip with Network.HTTP.Client, I'm relatively certain you need to decompress that response body you get back yourself

I'm skeptical because [http-client](https://hackage.haskell.org/package/http-client-0.7.16/docs/Network-HTTP-Client.html#v:requestHeaders) says

> the Accept-Encoding header is set implicitly to gzip for convenience by default

pbrisbin: I'm going to invent a Semigroup, aren't I?

```hs
instance MonadHttp (HandlerFor App) where
  httpLbs = httpGzip
    <> httpCached memcachedHttpCacheSettings
    <> httpRetry429s
    <> maybe httpStubbed (hoistIO httpLbs) appStubs
```

pbrisbin: Oh, how about that.

Heh, well now I'm back to -- do we need to rewrite `Network.HTTP.Client` to accomplish that feature?

library/Freckle/App/Http/Cache/Memcached.hs
chris-martin: If the purpose of what is now called `CResponse` is its `Serialise` instance, then yeah I think "serialise" in the name would be a good mnenonic. A comment that says "This type defines how to serialize a `CachedResponse`" would do a lot too.

library/Freckle/App/Http/Cache/Memcached.hs
chris-martin: Set `decompress = const False` on the `Request` to stop it from decompressing automatically? Then... maybe there's some stuff in `Network.HTTP.Client.Internal` that can be used but yeah I guess you'd be left with rewriting some of the unzipping logic there.

library/Freckle/App/Memcached/Client.hs
chris-martin: > which doesn't work

Wait, what? That's what we're doing in the session library https://github.com/freckle/yesod-session-persist/blob/757f7a7355d0afdab4b8ab39e335daafee7191d3/internal/Yesod/Session/Memcache/Storage.hs#L57

pbrisbin: Oh nice. You read the docs more thoroughly than I did.

https://github.com/freckle/yesod-session-persist/blob/757f7a7355d0afdab4b8ab39e335daafee7191d3/internal/Yesod/Session/Memcache/Storage.hs#L101-L112

This should be fine then!

chris-martin: Oh, do you mean "doesn't work" as in it doesn't have the appropriate behavior in this use case? Because you actually want the check-and-set here to make sure you're evicting the old value?

chris-martin: Hiding a get-and-delete within this function, in this case, defeats the purpose.

I'm confused why you need to evict at all. Why not just let the new fetch overwrite the old value?

pbrisbin: Nope, I was just mistaken. I didn't know `0` had specially meaning and I though this code would only actually delete something if it had the literal version 0, which I expected to not be right. Happy to find out it works differently.

pbrisbin: > I'm confused why you need to evict at all. Why not just let the new fetch overwrite the old value?

This is a good point. I'll admit that I didn't actually have a reason and the code just ended up that way.

But thinking now, there is one scenario where it matters:

1. GET => 200 Cache-Control max-age=5 ETag=x
2. 6 seconds pass, our cache is stale, we'll use ETag and hope for a 304
3. GET If-Non-Match: x => 200 Cache-Control no-store ETag=y

At step (3) we need to evict, but should not overwrite with anything new.

The test suite sets up this scenario, but not because I thought about it as A Thing To Handle, but because it was the only way to observe eviction at all.

---

Things would be a lot simpler if I didn't care about this case. It drops one whole operation out of the `HttpCache` record and simplifies the stair-`case`, so I'm wondering... should I not care?

pbrisbin:
OK, we are now caching in gzipped form when possible!

Very glad I implemented `Freckle.App.Test.Http` before attempting this PR. The test coverage of all this feels really sound.

chris-martin:
I'd just set `http-types-0.12.4` in the LTS 20 config to fix that build, it's a non-breaking change so should be fine to require it.

library/Freckle/App/Http/Cache.hs
```diff
@@ -0,0 +1,287 @@
...
+--
+-- If cacheable, the 'CacheKey' is built from: method, scheme, host, port, path,
+-- query + any @Vary@ headers.
+getCachableRequestKey
+  :: HttpCacheSettings m t -> Request -> Maybe CacheKey
+getCachableRequestKey settings req = do
+  guard $ settings.cacheable req
+  guard $ HTTP.method req == "GET"
+  guard $ NoStore `notElem` requestHeaders.cacheControl
+  guard $ not settings.shared || Private `notElem` requestHeaders.cacheControl
```
chris-martin, line 166: I love how much standard stuff we're getting from HTTP here, I've never appreciated HTTP as much as I do now.

pbrisbin: It's almost like it was well-designed! At least in this area...

pbrisbin:
> I'd just set http-types-0.12.4 in the LTS 20 config to fix that build

Yeah, I might. I already have CCP in the module though, so It's not a huge deal to guard the necessary orphan `Generic` instances. Looking up the version to guard on was really what I was procrastinating on, so you saved me there!

library/Freckle/App/Http/Cache.hs
```diff
@@ -0,0 +1,287 @@
...
+
+cacheControlMaxAge :: [CacheControl] -> Maybe Seconds
+cacheControlMaxAge = firstJust $ \case
+  MaxAge s -> Just s
+  _ -> Nothing
+
+readCacheControl :: ByteString -> Maybe CacheControl
+readCacheControl = \case
+  "private" -> Just Private
+  "no-store" -> Just NoStore
```
chris-martin, line 241: Technically, parsing cache-control directives should be case-insensitive

chris-martin (approved):
Ship it

PR #160, opened 2024-03-19
Introduce FaktoryProducerPool 

z0isch:
We should rely on a pool of connections to Faktory when  enqueueing jobs instead of opening new connections every time. This introduce a typeclass to abstract some of that.

### DIFF
--- a/freckle-app.cabal
+++ b/freckle-app.cabal
@@ -5,7 +5,7 @@ cabal-version: 1.18
 -- see: https://github.com/sol/hpack
 
 name:           freckle-app
-version:        1.15.1.0
+version:        1.15.2.0
 synopsis:       Haskell application toolkit used at Freckle
 description:    Please see README.md
 category:       Utils
--- a/freckle-app.cabal
+++ b/freckle-app.cabal
@@ -43,6 +43,7 @@ library
       Freckle.App.Exception.MonadThrow
       Freckle.App.Exception.MonadUnliftIO
       Freckle.App.Exception.Types
+      Freckle.App.Faktory.ProducerPool
       Freckle.App.Ghci
       Freckle.App.GlobalCache
       Freckle.App.Http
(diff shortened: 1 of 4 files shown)


freckle-app.cabal
```diff
@@ -140,6 +141,7 @@ library
     , errors
     , exceptions
     , extra
+    , faktory
```
pbrisbin, line 144: Ah, I didn't realize *none* of our existing Faktory wrapper stuff was here yet. Now I remember.

https://app.asana.com/0/1204746045920874/1200495324837436/f

Stuff like `HasFaktorySettings`, `enqueueSink` maybe, could all be brought in here.

pbrisbin (approved):
LGTM modulo the conflicts, and ensuring the version bump gets to `freckle-app.cabal`.

Sorry I forgot that we hadn't gotten the _existing_ Faktory stuff to a good place, so you were kind of set up to follow a bad path there.

freckle-app.cabal
z0isch: That was why I was hesitant to bring this over here yeah- it was pretty annoying to do the dev with it being split up, but I wasn't exactly sure which pieces we would want to move over. 

I guess anything that both `jobs` and `fancy-api` uses would make sense as a starting point.

pbrisbin: > I guess anything that both jobs and fancy-api uses would make sense as a starting point

Anything that's business-agnostic is fair game for freckle-app. That sharing is a good signal for prioritizing though; in case we ever want to pull out `jobs` code into something stream-aligned, it would be useful to have those bits extracted for sure.

pbrisbin: I do think all of this logic could live in `faktory` too, but there are some questions there. That library is simplistic in terms of being concretely `IO`. If we're going to start talking about `MonadReader env m, HasSomething env` within that library, we might want to consider migrate it to a more abstract monad (which would open up additional niceties like relying on `MonadLogger`, `MonadTrace`, etc).

PR #166, opened 2024-05-16
Don't be Datadog-specific in OpenTelemetry module

pbrisbin:
The existing OpenTelemetry modules here were careful to assume and
support Datadog as a backend. This was because we thought you had to in
order for it to function. That was either not the case, or is no longer
the case, as Datadog seems able to handle default id values and even
prefers to encode them as hex in its own UI. Converting ids to a
`Word64` value is no longer necessary, and actively confusing to see.

Because of this, we can simplify and remove some of our code here. This
change deprecates various things to nudge users a certain way when
upgrading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,49 @@
 ## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.15.3.0...main)
 
+-- | __DEPRECATED__
+--
+-- This module used to wrap
+-- 'OpenTelemetry.Instrumentation.Wai.newOpenTelemetryWaiMiddleware' with
+-- additional behaviors:
+--
+-- 1. Add an @X-Datadog-Trace-Id@ response header
+-- 2. Add a 'Word64'-encoded @trace_id@ to logging context
+--
+-- (1) should be handled (better) by using an actual @datadog@ propagator. That
+-- said, it's also not necessary to do. The default propagators add trace
+-- context as thier own headers, which are sufficient to use.
+--
+-- (2) doesn't need to be 'Word64'-encoded. It seems Datadog can now work with
+-- the more common 'Base16'-encoded (i.e. hex) values. This is what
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,26 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.15.3.0...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/v1.15.4.0...main)
+
+## [v1.15.4.0](https://github.com/freckle/freckle-app/compare/v1.15.3.0...v1.15.4.0)
+
+- Add `addThreadContextFromTracing` and deprecate existing middleware
+
+  Migrating away from the deprecated middleware would look like:
+
+  ```diff
```
z0isch, line 9: Thanks for the nice diff!

library/Freckle/App/OpenTelemetry/ThreadContext.hs
```diff
@@ -0,0 +1,53 @@
...
+import OpenTelemetry.Trace.Id
+  ( Base (..)
+  , spanIdBaseEncodedText
+  , traceIdBaseEncodedText
+  )
+import OpenTelemetry.Trace.TraceState (Key (..), TraceState (..), Value (..))
+
+-- | Add tracing context values to the logging context
+--
+-- Values are encoded to 'Base16' (e.g. hex).
```
z0isch, line 24: Nit: Do you mean i.e. here?

z0isch (approved):

library/Freckle/App/OpenTelemetry/ThreadContext.hs
pbrisbin: I do!

pbrisbin, line 24:
```suggestion
-- Values are encoded to 'Base16' (i.e. hex).
```

PR #167, opened 2024-05-22
Add Freckle.App.OpenTelemetry.Context

pbrisbin:
This will support propagating trace context from an instrumented service
to another via HTTP request or Jobs enqueue. It works by attaching our
trace context to request headers ("inject") when sending messages (HTTP
requests or Jobs payloads) and then reading those request headers when
processing requests (handled by the WAI middleware for web apps and
bespoke code for Jobs).

Since propagation works on request headers, we add a small `HasHeaders`
class (it works with request or response headers, so we don't qualify)
and then provide instances for actual `Request`s as well as `Job`
payloads, via its `custom` property. More instances could be added
later, such as for Kafka events, etc.


tests/Freckle/App/OpenTelemetry/ContextSpec.hs
```diff
@@ -0,0 +1,156 @@
...
+            `shouldBe` Just (spanIdToHex $ Trace.spanId spanContext)
+
+assertCurrentSpanContext :: (MonadIO m, HasCallStack) => m Trace.SpanContext
+assertCurrentSpanContext =
+  getCurrentSpanContext >>= \case
+    Nothing -> expectationFailure "Expect there to be a SpanContext"
+    Just spanContext -> pure spanContext
+
+toTraceParent :: Text -> Text -> Text
+toTraceParent traceId spanId = "00-" <> traceId <> "-" <> spanId <> "-00"
```
z0isch, line 149: What's up with needing the prefix and suffix?

library/Freckle/App/OpenTelemetry/Context.hs
```diff
@@ -0,0 +1,116 @@
...
+import OpenTelemetry.Context (Context)
+import OpenTelemetry.Context.ThreadLocal (attachContext, getContext)
+import OpenTelemetry.Propagator (Propagator, extract, inject)
+import OpenTelemetry.Trace.Core
+  ( getTracerProviderPropagators
+  , getTracerTracerProvider
+  )
+
+class HasHeaders a where
+  headersL :: Lens' a [Header]
```
z0isch, line 39: Feels a touch weird to be using `Network.HTTP.Types.Header` for things that may not be http. It's just a type alias, right? I might be less confused if we use the actual type: `(CI ByteString, ByteString)`

z0isch (approved):

tests/Freckle/App/OpenTelemetry/ContextSpec.hs
pbrisbin: This just just how OpenTelemetry formats the `traceparent` header it uses in context propagation. I'm sure we could look up the spec to see what the values mean. I'm sure `00` is just what I'm getting in test because there's no sampling to encode, or something like that.

PR #170, opened 2024-06-03
upgrade to Blammo 1.2, remove LoggingT from monad stacks

chris-martin:


library/Freckle/App.hs
```diff
@@ -225,21 +224,19 @@ newtype AppT app m a = AppT
     , MonadThrow
     , MonadCatch
     , MonadMask
-    , MonadLogger
-    , MonadLoggerIO
     , MonadResource
     , MonadReader app
     )
+  deriving (MonadLogger, MonadLoggerIO) via WithLogger app (ResourceT m)
```
pbrisbin, line 230: :ok_hand:

library/Freckle/App.hs
```diff
@@ -225,21 +224,19 @@ newtype AppT app m a = AppT
...
 
 instance PrimMonad m => PrimMonad (AppT app m) where
   type PrimState (AppT app m) = PrimState m
 
   -- This should really just be `lift . primitive`, but:
   --
   -- - We'd need `MonadTrans (AppT app)`, which meh
-  -- - We'd need an orphan `Primitive LoggingT`, which no thanks
```
pbrisbin, line 240: Yesss!

Maybe we should give `MonadTrans (AppT app)` now :thinking:

library/Freckle/App.hs
```diff
@@ -259,6 +256,6 @@ instance
   where
   runSqlTx = runDB
 
-runAppT :: (MonadUnliftIO m, HasLogger app) => AppT app m a -> app -> m a
+runAppT :: MonadUnliftIO m => AppT app m a -> app -> m a
```
pbrisbin, line 259: I'm surprised this removed a constraint. I guess what happens now is this is perfectly usable without `HasLogger env` but when you go an actually use a `logInfo` (or whatever), it'll error for no instance _then_ since it can't solve `MonadLogger` without it?

library/Freckle/App/Test.hs
```diff
@@ -106,12 +108,12 @@ instance PrimMonad (AppExample app) where
...
   primitive = liftIO . primitive
 
-instance HasLogger app => Example (AppExample app a) where
+instance Example (AppExample app a) where
   type Arg (AppExample app a) = app
 
   evaluateExample (AppExample ex) params action =
     evaluateExample
-      (action $ \app -> void $ runLoggerLoggingT app $ runReaderT ex app)
+      (action $ \app -> void $ runReaderT ex app)
```
pbrisbin, line 116: And here would fix our test performance regression to, eh?

pbrisbin (approved):

library/Freckle/App.hs
chris-martin: I don't see why not

library/Freckle/App/Test.hs
chris-martin: One of them, at least!

PR #172, opened 2024-06-06
update hw-kafka-client

chris-martin:
Fixes for two breaking changes in `hw-kafka-client`:

- The `produceMessageBatch` function [was removed](https://github.com/haskell-works/hw-kafka-client/pull/179). We now repeatedly call `produceMessage` instead. I'm not sure if this is a big deal or not; were we relying on the bulk operation for performance or semantics?

- `ProducerRecord` now has a `prHeaders` field. I'm not sure what this is for, but setting it to `mempty` seems to work.

This could set our `hw-kafka-client` lower bound to `4.0.4`, but since that was deprecated (perhaps because `4.0.3` -> `4.0.4` was a breaking change violating PVP), the version is now set to at least `5` in all of our Stack configurations.

Adding a bunch of reviewers because I don't know anything about Kafka and not sure who does.

### DIFF
--- a/library/Freckle/App/Kafka/Producer.hs
+++ b/library/Freckle/App/Kafka/Producer.hs
@@ -15,6 +15,8 @@ module Freckle.App.Kafka.Producer
 import Freckle.App.Prelude
 
 import Blammo.Logging
+import Conduit ((.|))
+import qualified Conduit
 import Control.Lens (Lens', view)
 import Data.Aeson (ToJSON, encode)
 import Data.ByteString.Lazy (toStrict)
@@ -91,13 +93,13 @@ createKafkaProducerPool
   -> KafkaProducerPoolConfig
   -> IO (Pool KafkaProducer)
 createKafkaProducerPool addresses KafkaProducerPoolConfig {..} =
-  Pool.newPool $
-    Pool.setNumStripes (Just kafkaProducerPoolConfigStripes) $
-      Pool.defaultPoolConfig
-        mkProducer
-        closeProducer
-        (realToFrac kafkaProducerPoolConfigIdleTimeout)
-        kafkaProducerPoolConfigSize
+  Pool.newPool
+    $ Pool.setNumStripes (Just kafkaProducerPoolConfigStripes)
+    $ Pool.defaultPoolConfig
+      mkProducer
+      closeProducer
+      (realToFrac kafkaProducerPoolConfigIdleTimeout)
+      kafkaProducerPoolConfigSize
  where
   mkProducer =
     either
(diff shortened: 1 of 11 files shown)


library/Freckle/App/Kafka/Producer.hs
```diff
@@ -123,25 +125,27 @@ produceKeyedOn prTopic values keyF = traced $ do
...
-        liftIO $
-          Pool.withResource producerPool $ \producer ->
-            produceMessageBatch producer $
-              toList $
-                mkProducerRecord <$> values
-      unless (null errors) $
-        logErrorNS "kafka" $
-          "Failed to send events" :# ["errors" .= fmap (tshow . snd) errors]
+      errors <- liftIO $ Pool.withResource producerPool $ \producer ->
+        Conduit.runConduit
```
z0isch, line 129: What's up with using conduit here instead of a plain `traverse`?

z0isch (commented):
I don't see anything conceptually wrong here, especially with this [comment](https://github.com/haskell-works/hw-kafka-client/issues/88#issuecomment-473751863) essentially saying that the underlying library obviates the need to really support batch producing anyway.

pbrisbin (approved):
Sounds good. I was a little nervous about losing the batch produce, so I followed the links. It seems [well justified](https://github.com/haskell-works/hw-kafka-client/issues/88#issuecomment-473751863).

> Single-message is only an API frontend, internally librdkafka is still batching requests to brokers. And sending a message is never a blocking operation: librdkafka just puts a message into its internal queue and releases control immediately.

I have a fuzzy memory of us having to hold `hw-kafka-client` back to avoid shared library issues on build, but I may be making that up. If backend can build its docker images with this in play, then SGTM.

library/Freckle/App/Kafka/Producer.hs
chris-martin: I find it easier to write.

z0isch:
> [well justified](https://github.com/haskell-works/hw-kafka-client/issues/88#issuecomment-473751863).
>

Funny for us to both get to the same comment 😃

chris-martin:
> If backend can build its docker images with this in play

I guess I'll try a megarepo PR against this change before merging here then.

library/Freckle/App/Kafka/Producer.hs
pbrisbin: I didn't even notice this. I read your blog about this, and I'm into it in spirit. Particulary when you're doing some version of concat-map-maybe like this. But I dunno, the alternative is a quite a bit more accessible IMO:

```hs
errors <- liftIO $ Pool.withResource producerPool $ \producer ->
  fmap catMaybes $ traverse (produceMessage producer . mkProducerRecord) values
```

Maybe not qualifiying the `Conduit.` would make me feel differently.

z0isch: The reason I ask is that my hackles go up a bit when I see conduit, and I assume that we needed it for some sort of constant space style processing. 

I agree that `fmap catMaybes . traverse` is a bit simpler for me to read at least.

chris-martin:
Is it desirable to still accumulate all the errors into a single log message, or should they just log one by one as the occur?

pbrisbin:
> Is it desirable to still accumulate all the errors into a single log message

I don't remember why it worked like that. I can see how it'd be simpler to just log it as it happens, and I can't imagine why that wouldn't be acceptable or even better.

library/Freckle/App/Kafka/Producer.hs
```diff
@@ -91,13 +91,13 @@ createKafkaProducerPool
...
-        closeProducer
-        (realToFrac kafkaProducerPoolConfigIdleTimeout)
-        kafkaProducerPoolConfigSize
+  Pool.newPool
+    $ Pool.setNumStripes (Just kafkaProducerPoolConfigStripes)
+    $ Pool.defaultPoolConfig
+      mkProducer
+      closeProducer
+      (realToFrac kafkaProducerPoolConfigIdleTimeout)
+      kafkaProducerPoolConfigSize
```
pbrisbin, line 100: Don't worry about changing this right now, but I reread a little bit on [the docs here](https://hackage.haskell.org/package/resource-pool-0.4.0.0/docs/Data-Pool.html#v:defaultPoolConfig) and I'm not sure we ever want to `setNumStripes`. The default is to match the current capabilities, which is useful to ensure threads don't fight over the same stripe when pulling a value out of the pool. I can't imagine not wanting this. I think it'd be better to pretend stripes don't exist in our own pool configurations types.

I think we got confused because the docs talk about how size / stripes  might result in configured size `N` being actually-sized `N - 1` due to integer division, so maybe it's something someone might care about, but I really don't think we ever should.

PR #173, opened 2024-06-12
Drop deprecated stuff

pbrisbin:
We've now completed the update cycle in our apps, so this stuff can all go.

While writing this PR our latest hspec-junit-formatter got into nightly and so our use of deprecated functions broke that build. I included the minimal change necessary to address that, but I plan to attack `Freckle.App.Test.Hspec.Runner` more later.


pbrisbin:
This is not ready.

Since our latest junit made it into nightly, I'd like to deal with that in order to land this.

chris-martin (approved):

PR #175, opened 2024-06-18
Add `httpBytes`

mjgpy3:
See our monorepo, PR 34769 for a refactor that would have benefited from this.

We probably could just call `httpLbs` directly but I figured since this module is trying to centralize HTTP we'd want it re-exported here.

### DIFF
--- a/library/Freckle/App/Http.hs
+++ b/library/Freckle/App/Http.hs
@@ -14,6 +14,9 @@ module Freckle.App.Http
   , HttpDecodeError (..)
   , httpDecode
 
+    -- * Raw response
+  , httpBytes
+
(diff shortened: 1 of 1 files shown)


pbrisbin (commented):
> We probably could just call httpLbs directly but I figured since this module is trying to centralize HTTP we'd want it re-exported here.

Sorry, can you say more about why it's valuable to export the name `httpBytes` when `httpLbs` is already there? I don't really follow.

mjgpy3:
@pbrisbin Whoops, my bad. I thought that `httpLbs` was just the one from `Network.HTTP.Simple` and we were hiding it.

Thanks!

PR #176, opened 2024-06-21
Add `Freckle.App.Http (setRequestMethod)`

mjgpy3:
**Why?**

As far as I know there are currently 2 ways to use a non-`GET` request method:
1. including method in the request string (e.g. `parseRequest_ $ "POST " <> url`)
2. importing lower-level HTTP module and setting the field directly (e.g. `req { HTTP.method = "POST"}`)

**2** breaks the centralization we're providing with this module by requiring consumers to import an internal thing.

Also both of these are stringly typed, producing a runtime error if a mistake is made, exs. forgotten space, or casing:

> Note that the request method must be provided as all capital letters.

^ from `http-conduit`'s docs.

**Note**

We could just re-export `Network.HTTP.Simple (setRequestMethod)`, however, it still has the stringly-typed problem (it takes a `ByteString`). `StdMethod` gives us nice type-safety and covers all our current (and I'd bet future) use-cases.

### DIFF
--- a/library/Freckle/App/Http.hs
+++ b/library/Freckle/App/Http.hs
@@ -32,6 +32,7 @@ module Freckle.App.Http
   , setRequestBodyJSON
   , setRequestBodyURLEncoded
   , setRequestCheckStatus
+  , setRequestMethod
   , setRequestPath
   , disableRequestDecompress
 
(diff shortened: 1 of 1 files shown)


pbrisbin (approved):
> We could just re-export Network.HTTP.Simple (setRequestMethod), however, it still has the stringly-typed problem

Haha, I have never once thought or cared about this problem. I just use that method and (I guess) always with canonical values like `"GET"`. We should open a PR to make that `setRequestMethod` do what we're doing here, IMO.

PR #177, opened 2024-06-21
Add `Average` monoid from our monorepo

mjgpy3:
`Average` is fairly widely-used in our monorepo. We want to move widely used combinators out of our core library and into `freckle-app`.

I initially wanted to replace this module with
[average](https://hackage.haskell.org/package/average) but our implementation is more powerful (enum instances, weighted average).

This imitates naming convention used for `Freckle.App.Exception.MonadUnliftIO`. I decided on `.Monoid.` instead of `.Semigroup.` because there are `Monoid` definitions within.

The monorepo has no tests for this module, directly...

I could reasonably see this as being a new hackage package. Happy to do that but it'll be more work.

### DIFF
--- a/library/Freckle/App/Monoid/Average.hs
+++ b/library/Freckle/App/Monoid/Average.hs
@@ -1,0 +1,143 @@
+module Freckle.App.Monoid.Average
+  ( Average (..)
+  , average
+  , average1
+  , averageDatum
+  , getAverage
+  , getAverageInt
+  , getRoundedAverage
+  , weightedAverage
+  , weightedAverages
+  , AverageableEnum
+  , averageableEnum
+  , getAverageableEnum
+  ) where
+
+import Freckle.App.Prelude
+
+import Freckle.App.Foldable1 (foldMap1')
+import Autodocodec qualified
+import Data.Aeson (ToJSON (..))
+import Data.Semigroup.Foldable (Foldable1)
+import Test.QuickCheck
+
+average :: (Foldable t, Fractional a) => t a -> Maybe a
+average = getAverage . foldMap' averageDatum
+
+average1 :: (Foldable1 t, Fractional a) => t a -> a
+average1 = getAverageSafe . foldMap1' averageDatum
+ where
+  getAverageSafe (Average l n) = n / fromIntegral l
+
+data Average n = Average !Int !n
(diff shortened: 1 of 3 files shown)


library/Freckle/App/Monoid/Average.hs
```diff
@@ -0,0 +1,143 @@
...
+instance (Arbitrary a, Enum a) => Arbitrary (AverageableEnum a) where
+  arbitrary = averageableEnum <$> arbitrary @a
+
+averageableEnum :: Enum a => a -> AverageableEnum a
+averageableEnum =
+  AverageableEnum . averageDatum . fromIntegral . (+ 1) . fromEnum
+
+getAverageableEnum :: Enum a => AverageableEnum a -> Maybe a
+getAverageableEnum (AverageableEnum avg) =
+  toEnum . subtract 1 <$> getRoundedAverage avg
```
pbrisbin, line 143: Is this really the sort of thing that should live in our general-purpose application toolkit? It seems really special case to me.

> I initially wanted to replace this module with
[average](https://hackage.haskell.org/package/average) but our implementation is more powerful (enum instances, weighted average)

If we left such single-use things with their single-uses, could you indeed move to `average`?

library/Freckle/App/Monoid/Average.hs
```diff
@@ -0,0 +1,143 @@
...
+--
+weightedAverages
+  :: (Integral weight, Foldable f, Num n) => f (n, weight) -> Average n
+weightedAverages correctnesses = foldMap avg correctnesses
+ where
+  avg (correctness, worth) = weightedAverage worth correctness
+
+-- Construct a weighted average with integral weights.
+--
+-- This is useful for calculating correctness values for assignment sessions.
```
pbrisbin, line 84: > calculating correctness values for assignment sessions

This feels business-domainey...

pbrisbin (commented):
>  because there are Monoid definitions within.

This is true of basically everything in `Data.Semigroup` too, so I'm not sure if that's how that split was decided. It seems like maybe `Data.Monoid` only contains backwards-compatibility newtypes?

mjgpy3:
Yeah, probably doesn't really belong here. I'll see what it'd take to get it in `average`.

PR #179, opened 2024-06-26
Add test assertion for `PersistValue` round-trip

mjgpy3:
Another assertion along the lines of JSON/PathPiece assertions. We test this occasionally in our monorepo.

### DIFF
--- a/library/Freckle/App/Test/Properties/PersistValue.hs
+++ b/library/Freckle/App/Test/Properties/PersistValue.hs
@@ -1,0 +1,11 @@
+module Freckle.App.Test.Properties.PersistValue
+  ( prop_roundTripPersistValue
+  ) where
+
+import Freckle.App.Prelude
+
+import Database.Persist (PersistField (..))
+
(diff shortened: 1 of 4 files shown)


library/Freckle/App/Test/Properties/PersistValue.hs
```diff
@@ -0,0 +1,11 @@
+module Freckle.App.Test.Properties.PersistValue
+  ( prop_roundTripPersistValue
+  ) where
+
+import Freckle.App.Prelude
+
+import Database.Persist (PersistField (..))
+
+-- | Check that @fromPersistValue (toPersistValue value)@ is @value@
```
pbrisbin, line 9:
```suggestion
-- | Check that @'fromPersistValue' ('toPersistValue' value)@ is @value@
```

pbrisbin (approved):

chris-martin:
Going to go ahead and push this in to empty out the outstanding PR list.

PR #181, opened 2024-07-19
move freckle-app into a subdirectory

chris-martin:
Same as with [Blammo](https://github.com/freckle/blammo), I think this repository could make a good launching pad for smaller packages. It obviously won't be quite as cohesive as the Blammo repository because our starting point is not very cohesive. We should probably have ambitions to throw some packages outside the bounds of this repository. I envision this repo holding a collection of packages I believe @pbrisbin described that still aim to serve as Freckle's foundations and a normalizing force across the org, just split up a bit more granularly because e.g. a lot of the foundations of a web service differ from those of a command-line app, etc.


pbrisbin (commented):
I'm a fan of everything you say in the PR description, my only concern is pushing things into `./freckle-app` before we have any _concrete_ plan for sprouting even one other thing next to it. I worry it puts the repo into an unexpected state and ends up left like that for an extended time.  Why do it now instead of waiting for a need to arise (and focusing our energies on defining it)?

chris-martin:
In my own projects I always start repositories with the package in a subdirectory even if there's only one, because... why not? I don't follow the "unexpected state" description, it seems perfectly ordinary.

pbrisbin (approved):

chris-martin:
A first step might be to do more or less the same thing as [Blammo-wai](https://github.com/freckle/blammo/pull/49). Create a new `freckle-web` package, move anything web-server-related into it?

PR #184, opened 2024-07-25
fix release workflow

chris-martin:
Flubbed the merge of #183 because I hadn't updated the release workflow to tag and release the new packages. I think this is an improvement?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,13 +7,20 @@ on:
 jobs:
   tag:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package:
+          - freckle-app
+          - freckle-env
+          - freckle-kafka
+
     steps:
       - uses: actions/checkout@v4
       - id: tag
         uses: freckle/haskell-tag-action@v1
         with:
-          package-yaml: freckle-app/package.yaml
-          tag-prefix: freckle-app-v
+          package-yaml: ${{ matrix.package }}/package.yaml
+          tag-prefix: ${{ matrix.package }}-v
     outputs:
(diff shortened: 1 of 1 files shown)


chris-martin:
Maybe I shouldn't have combined everything into one job, I just wasn't sure how to define two jobs that use the same matrix.

chris-martin:
... okay, redid this as a more minimal change.

.github/workflows/release.yml
```diff
@@ -38,7 +51,7 @@ jobs:
 
       # Upload using our oldest, tested resolver, to set the lowest lower bounds
       - name: "Release to Hackage"
-        run: stack --stack-yaml stack-lts-20.26.yaml upload --pvp-bounds lower freckle-app
+        run: stack --stack-yaml stack-lts-20.26.yaml upload --pvp-bounds lower ${{ matrix.package }}
 
       - name: "Upload documentation"
-        run: stack upload --documentation freckle-app
+        run: stack upload --documentation ${{ matrix.package }}
```
pbrisbin, line 57: This isn't normally necessary, it was added because Hackage can't build with the librdkafka dependency. That means I think it can be:

```suggestion
        if: ${{ matrix.package == 'freckle-kafka' }}
        run: stack upload --documentation ${{ matrix.package }}
```

Alternatively, you could add something like,

```yaml
include:
  - package: freckle-kafka
    upload-docs: true
```

Then make this `matrix.upload-docs == 'true'`. That would be useful if we think additional packages (but still not all) will need this step, but I think that's pretty unlikely.

pbrisbin (changes requested):

pbrisbin (approved):

PR #186, opened 2024-07-29
Add `TextualEnum` `newtype` wrapper

mjgpy3:
This is an experiment that I'm putting out for review to see what the team thinks. I won't feel bad if we don't like it.

**What is it?**

`TextualEnum` is a `newtype` wrapper meant to provide a lot of boilerplate serialization/deserialization instances that we keep defining and redefining as we introduce new types and services.

**How's it work?**

The core idea is that if you can give your type an `Enum` instance, a `Bounded` instance, and a way to convert to text, then that should be enough to provide string-type (i.e. textual) representations of it.

So `toText` tells us how to serialize the value, and we can look through the `Bounded`/`Enum` instance values to parse it "from text".

`toText` is preferable from `fromText` because, for bounded/enums, totality is straightforward to check/implement and further enforced if cases are updated/removed/added.

**EnumName**

I've also included `EnumName` because we've started introducing codec/schema definitions that can be named.

**Example**

You can see the specs (`PrimaryColor`) for an example.

### DIFF
--- a/freckle-app/library/Freckle/App/TextualEnum.hs
+++ b/freckle-app/library/Freckle/App/TextualEnum.hs
@@ -1,0 +1,116 @@
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | Typical instances for enumerated data types with textual representation
+module Freckle.App.TextualEnum
+  ( TextualEnum (..)
+  , EnumValue (..)
+  , EnumName (..)
+  , enums
+  , fromText
+  , prop_roundTripEnumText
+  ) where
+
+import Freckle.App.Prelude
+
+import Autodocodec
+import Control.Lens hiding (elements)
+import Control.Monad (mzero)
+import Data.Aeson
+import Data.Csv qualified as CSV
+import Data.Dynamic (Typeable)
+import Data.List.NonEmpty qualified as NE
+import Data.OpenApi
+import Data.Text.Encoding qualified as T
+import Database.Persist.Sql
+  ( PersistField (..)
+  , PersistFieldSql (..)
+  , SqlType (..)
+  )
+import Servant
+import Test.QuickCheck (Arbitrary (..), elements)
+import Web.PathPieces
+
+class EnumValue a where
+  -- | Convert a 'TextualEnum' to 'Text'
+  toText :: a -> Text
+
+class EnumName a where
+  -- | Name of a 'TextualEnum', used for naming schemas
+  enumName :: Proxy a -> Text
+
+-- | Wrapper around enums
+--
+-- N.B. This should not be used for "enormous" enumerations. It's primary purpose
+-- is to provide standard instances for discriminated-union-defined enums.
+newtype TextualEnum a = TextualEnum {enumValue :: a}
+  deriving newtype (Eq, Show, Ord, Generic)
+
+-- | All values of a 'TextualEnum'
+enums :: (Bounded a, Enum a) => NonEmpty (TextualEnum a)
+enums = fmap TextualEnum $ minBound NE.:| drop 1 [minBound .. maxBound]
+
+-- | Parse a 'TextualEnum' from 'Text'
+fromText :: (EnumValue a, Bounded a, Enum a) => Text -> Maybe (TextualEnum a)
+fromText txt = find ((== txt) . toText . enumValue) enums
+
+instance EnumValue a => ToJSON (TextualEnum a) where
+  toJSON = toJSON . toText . enumValue
+  toEncoding = toEncoding . toText . enumValue
+
+instance (EnumValue a, Bounded a, Enum a) => FromJSON (TextualEnum a) where
+  parseJSON = withText "TextualEnum" $ maybe mzero pure . fromText
+
+instance (Bounded a, Enum a) => Arbitrary (TextualEnum a) where
+  arbitrary = elements $ toList enums
+
+instance (EnumValue a, Bounded a, Enum a) => PathPiece (TextualEnum a) where
+  toPathPiece = toPathPiece . toText . enumValue
+  fromPathPiece = fromText
+
+instance EnumValue a => CSV.ToField (TextualEnum a) where
+  toField = CSV.toField . toText . enumValue
+
+instance (EnumValue a, Bounded a, Enum a) => CSV.FromField (TextualEnum a) where
+  parseField = maybe mzero pure . fromText . T.decodeUtf8
+
+instance
+  (Bounded a, Enum a, EnumValue a, Typeable a, EnumName a)
+  => ToSchema (TextualEnum a)
+  where
+  declareNamedSchema = pure . NamedSchema (Just $ enumName $ Proxy @a) . enumOptions
+   where
+    enumOptions _ =
+      mempty
+        & type_ ?~ OpenApiString
+        & enum_ ?~ (toJSON . toText . enumValue <$> toList (enums @a))
+        & example ?~ (toJSON . toText . enumValue $ NE.head (enums @a))
+
+instance (Bounded a, Enum a, EnumValue a) => ToParamSchema (TextualEnum a) where
+  toParamSchema _ =
+    mempty
+      & type_ ?~ OpenApiString
+      & enum_ ?~ (toJSON . toText . enumValue <$> toList (enums @a))
+      & example ?~ (toJSON . toText . enumValue $ NE.head (enums @a))
+
+instance (Bounded a, Enum a, EnumValue a, Eq a) => HasCodec (TextualEnum a) where
+  codec = stringConstCodec $ (id &&& (toText . enumValue)) <$> enums @a
+
+instance (Bounded a, Enum a, EnumValue a) => PersistField (TextualEnum a) where
+  toPersistValue = toPersistValue . toText . enumValue
+  fromPersistValue =
+    maybe (Left "Not member of enumeration") Right . fromText <=< fromPersistValue
+
+instance (Bounded a, Enum a, EnumValue a) => PersistFieldSql (TextualEnum a) where
+  sqlType _ = SqlString
+
+instance (Bounded a, Enum a, EnumValue a) => FromHttpApiData (TextualEnum a) where
+  parseUrlPiece =
+    maybe (Left "Not member of enumeration") Right . fromText
+
+instance EnumValue a => ToHttpApiData (TextualEnum a) where
+  toUrlPiece = toText . enumValue
+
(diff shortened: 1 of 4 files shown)


freckle-app/library/Freckle/App/TextualEnum.hs
```diff
@@ -0,0 +1,116 @@
...
+
+instance (Bounded a, Enum a, EnumValue a) => ToParamSchema (TextualEnum a) where
+  toParamSchema _ =
+    mempty
+      & type_ ?~ OpenApiString
+      & enum_ ?~ (toJSON . toText . enumValue <$> toList (enums @a))
+      & example ?~ (toJSON . toText . enumValue $ NE.head (enums @a))
+
+instance (Bounded a, Enum a, EnumValue a, Eq a) => HasCodec (TextualEnum a) where
+  codec = stringConstCodec $ (id &&& (toText . enumValue)) <$> enums @a
```
chris-martin, line 96: Can you actually derive instance this via `EnumValue`? As we just happened to be discussing [a minute ago](https://github.com/freckle/megarepo/pull/35303#discussion_r1695732975) there seems to be some problem with deriving `HasCodec`.

freckle-app/package.yaml
```diff
@@ -167,6 +168,7 @@ tests:
       - mtl
       - nonempty-containers
       - postgresql-simple
+      - servant-server
```
chris-martin, line 171: To avoid piling up dependencies it might make sense to make multiple `TextualEnum`-like newtypes in different packages and then use them like

```haskell
newtype Foo = Foo Text
  deriving Csv.ToField via CsvTextualEnum Foo
  deriving Servant.Whatever via ServantTextualEnum Foo
  -- etc.
```

The textual enum concept is something that basically ever component of our software is going to want, so it would be nice if that didn't always have to automatically incur a dependency on myriad packages (cassava, persistent, servant, autodocodec, quickcheck, ...) that cover a vast variety of unrelated domains.

freckle-app/library/Freckle/App/TextualEnum.hs
```diff
@@ -0,0 +1,116 @@
...
+  , PersistFieldSql (..)
+  , SqlType (..)
+  )
+import Servant
+import Test.QuickCheck (Arbitrary (..), elements)
+import Web.PathPieces
+
+class EnumValue a where
+  -- | Convert a 'TextualEnum' to 'Text'
+  toText :: a -> Text
```
chris-martin, line 35: A thought -

```haskell
data TextMapping a =
  TextMapping !(a -> Text) !(Text -> Maybe a)

class EnumValue a where
  textMapping :: TextMapping a

enumBoundedTextMapping :: (Enum a, Bounded a) => (a -> Text) -> TextMapping a
enumBoundedTextMapping f =
  let
    !m = Map.fromList $ (id &&& f) <$> [minBound..maxBound]
  in
    TextMapping f (flip (Map.lookup m))
```

```haskell
data ContentArea = Math | Ela
  deriving (...) via TextualMapping ContentArea

instance EnumValue ContentArea where
  textMapping = enumBoundedTextMapping $ \case
    Math -> "math"
    Ela -> "ela"
```

This approach does two things -

- Allows the user options other than enum/bounded to enumerate if they want
- Allows a logarithmic rather than linear lookup from a `Map` which is constructed only once (not 100% sure if that would require explicitly lifting each `textMapping` definition to the top level or not, would want a benchmark to test.)

mjgpy3: I thought about doing something like this but I came to the conclusion that basically
> not 100% sure if that would require explicitly lifting each textMapping definition to the top level or not

was kind of hazy (i.e. I'm unsure exactly how to get GHC to compute it once) and thus prone to being incorrectly done.

I'm also not convinced that a hash map is actually quicker for smaller enums. Hashing a value vs. looking through like 4-13 cases feels like a wash. I could be totally wrong here as I have no data to back this up but the tradeoffs don't feel worth it at first blush.

freckle-app/package.yaml
mjgpy3: Maybe a good thing to do eventually, for sure.

freckle-app/library/Freckle/App/TextualEnum.hs
mjgpy3: Not sure I follow what you're asking. Are you asking if this instance is broken or if there's a simpler way to derive it?

freckle-app/library/Freckle/App/TextualEnum.hs
joelmccracken: personally I greatly prefer using deriving via instead of wrapping a bunch of types with a utility type; I'm fine with wrapping in a utility type for calculation purposes, but it gets really stinky when someone does e.g.

```
data Person = { age :: Sum Int } 

totalAge :: [Person] -> Int
totalAge = getSum .  fold . age 
```
instead of, like:
```
data Person = { age :: Int } 

totalAge :: [Person] -> Int
totalAge = getSum . fold <$> Sum . age 

-- which would let you then do

youngestAge :: [Person] -> Int
youngestAge = getMin . fold <$> Min . age 
```
.. tho I admit ive seen it in the our codebase a few times, and that this is a personal preference that I can't justify beyond "confuses the data model" and "at what point do you stop adding utility wrappers"

Tho, lol, I guess this utility wrapper thing is kinda sorta analogous to monad transformer stacks, eh?

I don't really have an opinion on the linear vs log search. If the N is small it doesn't matter much. I guess if someone did the following, it would matter a lot:

```
newtype Age = Age Int
  deriving (...) via TextualMapping Age

instance EnumValue ContentArea where
  textMapping = enumBoundedTextMapping show

main = do putStrLn $ show $ Age maxBound
```
but idk this seems like quite the goofy thing to do, and hey, I notice that you address it. If you do it via mapping, and someone were to enumerate the set of Ints, you'd end up eating up so much memory... so idk. 

you could make a way for the user to use a second function tho so a body could pick, e.g. `enumBoundedTextMapping` and `enumBoundedTextMappingMemo` or something

freckle-app/library/Freckle/App/TextualEnum.hs
chris-martin: Referring to what we mentioned in backend guild today that `HasCodec` for whatever reason isn't a class that works with deriving via. Just wondering how much that limits the utility of this instance.

freckle-app/library/Freckle/App/TextualEnum.hs
chris-martin: > greatly prefer using deriving via instead of wrapping a bunch of types with a utility type

Agreed and I'm wondering if maybe the spec here would be a good place to demonstrate using `TextualEnum` to define instances directly on the `PrimaryColor` type.

freckle-app/library/Freckle/App/TextualEnum.hs
joelmccracken: I think he's askjing if you can remove this explicit instance and use deriving via to achieve this

joelmccracken (changes requested):
lgtm but i'd like to resolve the "how this is to be used" (hopefully, not wrapping a bunch of domain types, but deriving via/newtype)

freckle-app/package.yaml
mjgpy3: @chris-martin, @joelmccracken  last commit has spec use deriving-via instead of newtype.

freckle-app/tests/Freckle/App/TextualEnumSpec.hs
```diff
@@ -0,0 +1,95 @@
...
+    , Servant.FromHttpApiData
+    , Servant.ToHttpApiData
+    , Path.PathPiece
+    , Sql.PersistField
+    , Arbitrary
+    )
+    via TextualEnum PrimaryColor
+
+instance EnumValue PrimaryColor where
+  toText = \case
```
joelmccracken, line 39: What about an example showing using https://hackage.haskell.org/package/base-4.20.0.1/docs/Data-Data.html#v:toConstr which would help DRY/standardize/make one less spot to forget to update when adding an enum value

chris-martin: I'm :-1: in general on things that couple strings to Haskell identifiers, because it makes the code less robust against refactoring, leads to compromising what strings should be to deal with Haskell restrictions (e.g. wanting an enum string to be "type"), and makes it harder to search the code when you're expecting to be able to find a string literal appearing somewhere but it's actually derived from a Haskell identifier instead -- But sometimes I guess it's appropriate, and TIL about `toConstr`, so wouldn't argue against more docs.

freckle-app/library/Freckle/App/TextualEnum.hs
chris-martin: I'm saying I know you cannot use deriving via to achieve this, so I don't understand what purpose the instance serves.

freckle-app/library/Freckle/App/TextualEnum.hs
```diff
@@ -0,0 +1,116 @@
...
+
+class EnumName a where
+  -- | Name of a 'TextualEnum', used for naming schemas
+  enumName :: Proxy a -> Text
```
chris-martin, line 39: How about making this derivable?

```haskell
newtype DataName a = DataName a

instance Data a => EnumName (DataName a) where
  enumName _ = _ -- Some sort of reflection?

data MyThing = ...
  deriving EnumName via DataName MyThing
```

chris-martin (approved):
A few ideas for additional features floating around but I think this will be useful.

freckle-app/tests/Freckle/App/TextualEnumSpec.hs
joelmccracken: 🤷 I get that pov. I do go back and forth. so much is automatically generated, and depending upon tradeoffs, it becomes unclear when specifically someone would be searching for a string vs somewhere else. But yea if you do autogen these, it really important that someone be able to decouple these when its needed, and sometimes stuff don't have "i need to roll my own" docs. 

either way perhaps its worth documenting/commenting/illustrating using it? i'm good with either way.

mjgpy3:
@joelmccracken re-requesting a review on this. I _think_ I've covered your concerns but it's been a while so just wanted to make sure (sorry, I went on vacation shortly after opening this so I forgot about it).

joelmccracken (approved):
looks fine, honestly I barely remember any of this lol

PR #189, opened 2024-08-12
split freckle-(prelude,exception)

chris-martin:
As discussed in https://github.com/freckle/freckle-app/pull/185#discussion_r1695804089, the prelude extraction has the potential to reduce dependency trees of packages that only need `freckle-app` for the prelude.

I don't think the growing list of freckle-app module re-exports should stick around forever, but it's a nice way to ease migration.

Also including a little cleanup to docs that I was lazy about in previous package refactors.


mjgpy3 (approved):
Seems fine to me.

PR #191, opened 2024-09-05
update Freckle.App.Env to show the needful

joelmccracken:
The example dun work

### DIFF
--- a/freckle-env/library/Freckle/App/Env.hs
+++ b/freckle-env/library/Freckle/App/Env.hs
@@ -13,7 +13,7 @@
 -- >   }
 -- >
 -- > loadConfig :: IO Config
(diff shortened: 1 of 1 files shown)


freckle-env/library/Freckle/App/Env.hs
```diff
@@ -13,7 +13,7 @@
 -- >   }
 -- >
 -- > loadConfig :: IO Config
--- > loadConfig = parse $ Config
+-- > loadConfig = parse (header "your-program-name") $ Config
```
pbrisbin, line 16: You could also do `id` if you're going for minimal.
```suggestion
-- > loadConfig = parse id $ Config
```

pbrisbin (approved):
:saluting_face:

chris-martin (approved):

PR #192, opened 2024-09-18
runDB: add otel span for holding a database connection

chris-martin:
Curious what y'all think of this idea - I'd like our traces to more explicitly show within `runDB` spans how much time is spent

- waiting on the connection pool; versus
- actually holding a connection.

The nicest way to display this would be to have a "waiting on the connection pool" span, but since the pool acquire is done by `persistent`, that doesn't seem super easy. Putting a span around the action, though, is trivial.

- Pro: When the reason for an action being slow was a long wait on the connection pool, that will appear clearly in its trace.
- Con: Maybe clutters up the traces a bit, since it adds a level of depth under every `runDB` which won't be super informative under normal performance.

### DIFF
--- a/freckle-app/library/Freckle/App/Database.hs
+++ b/freckle-app/library/Freckle/App/Database.hs
@@ -176,7 +176,7 @@ runDB action = do
   Stats.withGauge Stats.dbEnqueuedAndProcessing $
     inSpan "runDB" (clientSpanArguments {Trace.attributes = dbAttributes}) $
       runSqlPoolWithExtensibleHooks
-        (inSpan "with DB connection" defaultSpanArguments action)
+        (inSpan "runSqlPool" defaultSpanArguments action)
         pool
         Nothing
         hooks'
--- a/freckle-app/library/Freckle/App/Database.hs
+++ b/freckle-app/library/Freckle/App/Database.hs
@@ -175,7 +175,11 @@ runDB action = do
         }
   Stats.withGauge Stats.dbEnqueuedAndProcessing $
     inSpan "runDB" (clientSpanArguments {Trace.attributes = dbAttributes}) $
-      runSqlPoolWithExtensibleHooks action pool Nothing hooks'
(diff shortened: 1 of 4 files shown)


freckle-app/library/Freckle/App/Database.hs
```diff
@@ -175,7 +175,11 @@ runDB action = do
         }
   Stats.withGauge Stats.dbEnqueuedAndProcessing $
     inSpan "runDB" (clientSpanArguments {Trace.attributes = dbAttributes}) $
-      runSqlPoolWithExtensibleHooks action pool Nothing hooks'
+      runSqlPoolWithExtensibleHooks
+        (inSpan "with DB connection" defaultSpanArguments action)
```
pbrisbin, line 179: I'd call the span `runSqlPool`.

pbrisbin (approved):
I have no issue with this, I just don't know that it's useful.

Today, the case we care about looks like this:

```
[ Handler                       ]
  [ runDB                     ]
                 [SELECT  ]

   ^^^^^^^^^^^^^^
   | dead time waiting
```

And with this change, I think it'd look like this:


```
[ Handler                       ]
  [ runDB                     ]
                [with DB co]
                 [SELECT  ]

   ^^^^^^^^^^^^^
   | dead time waiting
```

Does that teach us anything?

I guess if the traces come out _not_ looking like that, it'd teach us something.

chris-martin:
> I guess if the traces come out _not_ looking like that, it'd teach us something.

Yeah, it would tell us that the thread is doing something within `runDB` that it shouldn't be, which has been the theme of problems I've been trying to hunt down lately. If there are gaps _between_ or _after_ queries under the `runDB` (and there sometimes are) then you can see that something is wrong that's the fault of the code responsible for the trace you're looking at. If there is a gap at the _beginning_ of a `runDB`, you can't currently distinguish between

1. a fault lies in the code that was traced;
2. an exhausted connection pool.

danroyo (approved):

PR #193, opened 2024-09-18
add HasCallStack to a few yesod test utilities that call throwString

chris-martin:


freckle-app/library/Freckle/App/Test/Yesod.hs
```diff
@@ -159,20 +159,24 @@ testSetCookie :: forall m site. MonadYesodExample site m => SetCookie -> m ()
 testSetCookie = liftYesodExample . Yesod.Test.testSetCookie
 
 -- | Get the body of the most recent response and decode it as JSON
-getJsonBody :: forall a m site. (MonadYesodExample site m, FromJSON a) => m a
+getJsonBody
+  :: forall a m site. (MonadYesodExample site m, FromJSON a, HasCallStack) => m a
 getJsonBody = either err pure . eitherDecode =<< getRawBody
  where
   err e = throwString $ "Error decoding JSON response body: " <> e
```
pbrisbin, line 166: Drive-by, but I think it'd be nice if these used `expectationFailure` instead of `throwString`.

pbrisbin (approved):

freckle-app/library/Freckle/App/Test/Yesod.hs
chris-martin: Thank you, I think this might actually just be the solution to the problem I have (not seeing useful output when a test fails from this throw).

PR #194, opened 2024-09-18
add unwrapAnnotatedHUnitFailure

### LINKED ISSUE #190: Add utility for unwrapping AnnotatedException HUnitFailure
Here's a utility that every `SpecHook` is going to want:

```haskell
import Control.Exception qualified
import Control.Exception.Annotated (AnnotatedException (..))
import Test.HUnit.Lang (HUnitFailure)
import Test.Hspec

unwrapAnnotatedHUnitFailure :: Spec -> Spec
unwrapAnnotatedHUnitFailure = around_ $
  Control.Exception.handle $ \AnnotatedException {exception = e} ->
    Control.Exception.throw (e :: HUnitFailure)
```

This lets `hspec` catch and pretty-print `HUnitFailure`. Without it, if you end up with an `AnnotatedException HUnitFailure`, hspec doesn't recognize it and you get ugly output.

One of these packages (probably a testing library, but `freckle-app` for now since that isn't broken out yet) should include this function.

chris-martin:
closes #190

### DIFF
--- a/freckle-app/CHANGELOG.md
+++ b/freckle-app/CHANGELOG.md
@@ -1,11 +1,16 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-app-v1.20.0.2...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-app-v1.20.1.0...main)
 
-## [v1.20.0.2](https://github.com/freckle/freckle-app/compare/freckle-app-v1.20.0.1...freckle-app-v1.20.0.2)
+## [v1.20.1.0](https://github.com/freckle/freckle-app/compare/freckle-app-v1.20.0.1...freckle-app-v1.20.1.0)
+
+Modules moving between packages:
 
 - `Freckle.App.Prelude` has moved to the `freckle-prelude` package
 - `Freckle.App.Exception(.*)` has moved to the `freckle-exception` package
+- All moved modules are re-exported from `freckle-app`.
 
-All moved modules are re-exported from `freckle-app`.
+- Added `Freckle.App.Test.Hspec.AnnotatedException` (`unwrapAnnotatedHUnitFailure`).
+  It is recommended that all hspec test suites for packages that use `freckle-prelude`
+  add this spec hook.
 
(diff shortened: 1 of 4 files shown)


freckle-app/library/Freckle/App/Test/Hspec/AnnotatedException.hs
```diff
@@ -0,0 +1,22 @@
+module Freckle.App.Test.Hspec.AnnotatedException
+  ( unwrapAnnotatedHUnitFailure
```
pbrisbin, line 2: I might be wrong, but I _think the_ convention for libraries providing hooks is to name the module appropriately and expose only `hook`, then expect users to use it qualified.

```suggestion
module Freckle.App.Test.Hspec.Hooks.UnwrapAnnotatedException
  ( hook
```

:shrug: 

One concrete benefit would be the case where this is the only hook they need, you'd only need to re-export it:

```hs
module SpecHook
  ( module Freckle.App.Test.Hspec.Hooks.UnwrapAnnotatedException
  ) where

import Freckle.App.Test.Hspec.Hooks.UnwrapAnnotatedException
```

How common that is, I do not know.

pbrisbin (approved):

freckle-app/library/Freckle/App/Test/Hspec/AnnotatedException.hs
chris-martin: I can't find anything on Hackage's Hoogle resembling this pattern.

freckle-app/CHANGELOG.md
```diff
@@ -1,11 +1,16 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-app-v1.20.0.2...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-app-v1.20.1.0...main)
 
-## [v1.20.0.2](https://github.com/freckle/freckle-app/compare/freckle-app-v1.20.0.1...freckle-app-v1.20.0.2)
```
chris-martin, line 3: I apparently never actually made the version bump to release `1.20.0.2`, so those changes are going into `1.20.1.0` now.

Issue #159, opened 2024-03-14
Release Job needs a re-run to succeed


pbrisbin:
Currently, our haddocks fail to build on the first run of the `release` Workflow. It succeeds on re-run. For more details, and to track the eventual solution, see https://github.com/commercialhaskell/stack/issues/6270

pbrisbin:
Fixed in https://github.com/commercialhaskell/stack/commit/a3ad79447a48ac7fb07610285ef8bd80564c6ecd, not yet released.

chris-martin:
(I believe this is only the case for `freckle-kafka` releases now, since that's the only package we build haddocks for.)

pbrisbin:
This fix is present in Stack v3.1.1, which seems to be [what runners get now](https://github.com/actions/runner-images/blob/ubuntu22/20240825.1/images/ubuntu/Ubuntu2204-Readme.md#haskell-tools)\*. I expect our next release to not need this rerun.

\*There was a time GitHub runners lost the pre-installed `stack`, but I guess it's back.

PR #197, opened 2024-09-19
always upload haddocks

chris-martin:
We're not getting `freckle-app` haddocks on Hackage because the libpq dependency is making it fail to compile. If #159 really is fixed then I don't see the harm in just always uploading the docs.


pbrisbin (commented):
I'm not necessarily against always uploading manual haddocks, but it is quite a long process. Moving to the `include` style seems pretty trivial to retain only doing it when needed. I drafted that here so you could see what it looks like: https://github.com/freckle/freckle-app/pull/198. If you agree, we can merge that one instead.

PR #198, opened 2024-09-19
Use matrix.include to decide if haddocks need uploading

pbrisbin:
Originally, I thought only `freckle-kafka` needed this, but since
`freckle-app` is still depending on everything, nothing about its
system needs have changed, so it too needs haddocks manually built
(still).

Now that we have two packages like this, we can move to an
`include`-style, which is much easier to maintain.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,6 +46,14 @@ jobs:
           - freckle-otel
           - freckle-prelude
 
+        include:
+          # Packages with system dependencies (librdkafka, libpq) need haddocks
+          # built and uploaded manually.
+          - package: freckle-kafka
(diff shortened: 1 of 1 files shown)


chris-martin (approved):

chris-martin:
It seems like an extra list to maintain for dubious benefit; I'm not sure it's really a problem to spend a little extra GHA time building the docs since it doesn't block anything else. But I'm okay either way.

pbrisbin:
Yeah, it seems to take about 30m, which I might not call "a little", and this definitely doesn't scale to larger projects (backend would take well over an hour). But you're right that it doesn't block anything since the release has already happened at that point. So I agree, making it non-conditional is simpler.

PR #202, opened 2024-09-20
improve exceptions rethrown by unwrapAnnotatedHUnitFailure

### LINKED ISSUE #201: unwrapAnnotatedHUnitFailure should show annotations
I realize when trying to apply `unwrapAnnotatedHUnitFailure` in our app that it actually makes the output mostly worse. It does make [ExpectedButGot](https://hackage.haskell.org/package/HUnit-1.6.2.0/docs/Test-HUnit-Lang.html#v:ExpectedButGot) diffs display well, but at the cost of discarding annotations, which can be useful. Really what this hook should be doing is rendering the annotations and incorporating them into the `HUnitFailure` before re-throwing it unwrapped.

chris-martin:
This replicates a bunch of display logic from `displayException` in [annotated-exception](https://hackage.haskell.org/package/annotated-exception) that ideally would be exported by that package (cc @parsonsmatt).

Closes #201

### DIFF
--- a/freckle-app/library/Freckle/App/Test/Hspec/AnnotatedException.hs
+++ b/freckle-app/library/Freckle/App/Test/Hspec/AnnotatedException.hs
@@ -1,12 +1,18 @@
 module Freckle.App.Test.Hspec.AnnotatedException
   ( unwrapAnnotatedHUnitFailure
+  , annotateHUnitFailure
   ) where
 
 import Freckle.App.Prelude
 
 import Control.Exception qualified
+import Control.Lens (Lens', lens, over)
+import Data.Annotation (Annotation, tryAnnotations)
+import Data.List.NonEmpty (NonEmpty ((:|)), nonEmpty)
+import Data.Text qualified as T
 import Freckle.App.Exception (AnnotatedException (..))
-import Test.HUnit.Lang (HUnitFailure)
+import GHC.Stack (CallStack, prettyCallStack)
+import Test.HUnit.Lang (FailureReason (..), HUnitFailure (..))
 import Test.Hspec
 
 -- | An hspec hook that lets hspec catch and pretty-print 'HUnitFailure', the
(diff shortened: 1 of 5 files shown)


freckle-app/library/Freckle/App/Test/Hspec/AnnotatedException.hs
```diff
@@ -17,6 +23,85 @@ import Test.Hspec
...
+makeMessageMaybe mm as =
+  fmap combineParagraphs $
+    nonEmpty $
+      fmap stringParagraph (toList mm) <> annotationParagraphs as
+
+-- | Text that constitutes a paragraph in a potentially lengthy error message
+--
+-- Construct with 'stringParagraph' or 'textParagraph', which strip the text of
+-- surrounding whitespace.
+newtype Paragraph = Paragraph {paragraphText :: Text}
```
pbrisbin, line 68: This is all copied code? If so, OK. If we're designing this ourselves, I wonder if we want to use a real pretty-print library to build and then layout a `Doc` type.

freckle-app/tests/Freckle/App/Test/Hspec/AnnotatedExceptionSpec.hs
```diff
@@ -0,0 +1,147 @@
...
+            { annotations = [toAnnotation @Int 56]
+            , exception = HUnitFailure Nothing (Reason "x")
+            }
+          `shouldBe` HUnitFailure
+            Nothing
+            ( Reason
+                "x\n\
+                \\n\
+                \Annotations:\n\
+                \\t * Annotation @Int 56"
```
pbrisbin, line 44: FWIW, I find the literal newlines and continuation escapes make this hard to scan. Howe about something like,

```hs
Reason $ unlines
  [ "x"
  , ""
  , "Annotations:"
  , "    t * Annotation @Int 56"
  ]
```

Does it need to be easy to scan? Meh, not really. But I think a bug in the implementation string building, that was just offset by a bug in _this_ string building, would be hard to spot. I also just like, when testing string building code, to actually be able to understand the built string in test.

pbrisbin (approved):
I do wonder if the pain we self-inflict by such pervasive use of annotated exceptions will ever end. I do really like them though. :sweat_smile:

pbrisbin:
Wait, can we back up a bit as to why we need this hook at all?

The reason we need to be careful _in other scenarios_ is because we've done an `AnnotatedException.handle` + re-`throw` (which I think `checkpointCallstack` qualifies as). When we do that, we risk handling some else's exception (like `HandlerContents`) and re-`throw`-ing it as `AnnotatedException HandlerContents`, which breaks things, such as Yesod looking for that exception type to render a 400 or 404.

I can't see how our hspec usages would ever run afoul of this with `HUnitFailure`. As long as we are using hspec assertions (`shouldBe`) and not `AnnotatedException.throw HUnitFailure`, provided we don't have the handle-rethrow anywhere, and provided we don't manually checkpoint callstacks _around_ any assertions, there should never be an `AnnotatedException HUnitFailure` seen. I know those assumptions aren't guarantees, but they seem pretty stable to me in typical usage.

You originally got here because the `throwString` in `getJsonBody` was printing out the `AnnotatedException` instead of looking like an `HUnitFailure` -- and you thought that was due to excessive wrapping, right? But it wasn't. It was just throwing an annotated exception (by using our custom `throwString`) that was coming out as it was. Making that `expectationFailure` (or using `Control.Exception.throw HUnitFailure`, which is all that is) is the fix -- with or without a hook like this.

Am I misunderstanding something?

chris-martin:
I _originally_ got on this track because `shouldBe` failures should print like this:

```haskell
ghci> :m Test.Hspec Freckle.App.Prelude Freckle.App.Exception

ghci> hspec $ it "" $ 'a' `shouldBe` 'b'

Ghci2[7:9] [✘]

Failures:

  <interactive>:7:21: 
  1) Ghci2[7:9]
       expected: 'b'
        but got: 'a'

  To rerun use: --match "/Ghci2[7:9]/"

Randomized with seed 603656257

Finished in 0.0020 seconds
1 example, 1 failure
*** Exception: ExitFailure 1
```

But when they're annotated, they print like this:

```haskell
ghci> hspec $ it "" $ checkpointCallStack $ 'a' `shouldBe` 'b'

Ghci2[12:9] [✘]

Failures:

  <interactive>:12:9: 
  1) Ghci2[12:9]
       uncaught exception: AnnotatedException
       AnnotatedException {annotations = [Annotation @CallStack [("checkpointCallStack",SrcLoc {srcLocPackage = "interactive", srcLocModule = "Ghci2", srcLocFile = "<interactive>", srcLocStartLine = 12, srcLocStartCol = 17, srcLocEndLine = 12, srcLocEndCol = 36})]], exception = HUnitFailure (Just (SrcLoc {srcLocPackage = "interactive", srcLocModule = "Ghci2", srcLocFile = "<interactive>", srcLocStartLine = 12, srcLocStartCol = 43, srcLocEndLine = 12, srcLocEndCol = 53})) (ExpectedButGot Nothing "'b'" "'a'")}

  To rerun use: --match "/Ghci2[12:9]/"

Randomized with seed 1536058360

Finished in 0.0002 seconds
1 example, 1 failure
*** Exception: ExitFailure 1
```

> there should never be an `AnnotatedException HUnitFailure` seen

I don't remember why, but there was, it's a real problem that I hit in curricula. I don't think it's unusual.

- Hspec expectations can be anything with an `Example` instance, e.g. some custom `App` monad, and I don't think it's unreasonable to include a `checkpointCallStack` somewhere in the evaluation the application monad.
- Hspec assertions may happen within callbacks to some application-level thing that includes a `checkpointCallStack`.
- It is *good* for `HUnitFailure`s to get annotated with a call stack. In more complex tests, like when the assertion comes from some shared testing utility rather than from somewhere directly within the lexical scope of the spec that failed, a call stack is super helpful in figuring out what happened. (You can argue that we just shouldn't write tests this way, but we do.)

The only way to get a stack is to `checkpointCallStack`, and the only way to get a pretty diff printing is to convert `AnnotatedException` back to `HUnitFailure` right before it's handed off to the framework for formatting, hence this hook that combines the annotations with the HUnit failure to produce all the information we want in the format we want.

parsonsmatt:
Note that the new version of `annotated-exception` has a much improved `Show` instance - that alone may help solve some of the pain here

I'd be happy to accept a PR exposing anything you needed to copy out

freckle-app/library/Freckle/App/Test/Hspec/AnnotatedException.hs
chris-martin: Copied and modified. Pretty-print sounds good. #203

pbrisbin:
Assertions happening within your application monad (`AppExample`) or within an application callback that may be checkpointed is what I was missing.

PR #217, opened 2024-12-20
Add kafka batch consumer

z0isch:
Adds a new `runConsumerBatched` so that we can start playing with batched kafka consumers to see if it can help throughput.


OlaoluwaM (approved):
Awesome!

PR #222, opened 2025-01-07
remove persistent-sql-lifted

chris-martin:
It is moving; see https://github.com/freckle/persistent-sql-lifted/pull/1 - Neither PR blocks the other.

Also updating the nix flake while I'm here.


stack-nightly.yaml
```diff
@@ -31,7 +32,7 @@ extra-deps:
 
 flags:
   # https://github.com/haskell/directory/issues/176
-  directory: {os-string: true}
+  directory: { os-string: true }
```
pbrisbin, line 35: Huh, I wonder why Restyled didn't get me when I... oh, I think we need to migrate this repo to `.github/workflows/restyled.yml` still.

pbrisbin (approved):
Cool

PR #226, opened 2025-01-17
freckle-stats, freckle-ecs

chris-martin:
Splits outs two packages. 

- The statsd stuff is used by anything in production that does SQL queries to report its connection pool usage, and so this is one prerequisite for getting e.g. `curricula-api` off of the `freckle-app` monolith dependency.

- The ECS package is also pulled out at this time because `freckle-stats` needs it.

Fixes release action (stack config filename hadn't been updated)

Breaking change: A trivial Yesod/StatsD integration `instance HasStatsClient site => HasStatsClient (HandlerData child site)` has been removed and will need to be copied into individual applications.


freckle-stats/freckle-stats.cabal
```diff
@@ -0,0 +1,66 @@
+cabal-version: 1.18
+
+-- This file has been generated from package.yaml by hpack version 0.37.0.
+--
+-- see: https://github.com/sol/hpack
+
+name:           freckle-stats
```
pbrisbin, line 7: I think it's worth thinking about naming these more general-purpose extractions _not_ `freckle-`. Had we built them in isolation today, I can't imagine we would have; I think we're only are doing so now because we know its coming out of `freckle-app`, which isn't a very compelling reason to decide something as important as a name.

When this one was originally written, its purpose was to add `HasStatsClient`, then you'd use it has `MonadReader env m, HasStatsClient env`, so that's what it still is. I'm not sure I have a great name for that bit of functionality.

Again though, had we built this today, we would've most likely implemented a `MonadStats` as the central type of the package, with the pattern of `MockStatsClient` and `ActualStatsClient` types for deriving-via (the latter requiring that reader with `HasStatsClient`). If that were the case, `monad-stats` or `stats-mtl` are two strong naming conventions for packages that introduce a `MonadX` interface + functionality around it (e.g. `amazonka-mtl`).

I'm not sure what to do now though. Do we name it `stats-mtl` now and seek to move in that direction? Do we keep it `freckle-stats` and rename it if/when we move in that direction? Do we not move in that direction but come up with some other name?

pbrisbin (approved):

freckle-stats/freckle-stats.cabal
chris-martin: My feelings are that `freckle-` package names are expedients that are probably doomed to deprecation if/when they get turned into something that is better thought out for non-freckle users.

PR #241, opened 2025-06-03
add shouldBeJust

chris-martin:
I don't know what's the obsession with making assertions that return `()`; we are perpetually wanting to get the value out of a `Just` in tests.

### DIFF
--- a/freckle-app/CHANGELOG.md
+++ b/freckle-app/CHANGELOG.md
@@ -1,4 +1,8 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-app-v1.23.0.1...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-app-v1.23.2.0...main)
+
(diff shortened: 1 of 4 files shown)


pbrisbin (changes requested):
I think the reason all `should` functions return `()` is because that's the type of `it`. I would be very surprised and a bit annoyed to find that

```hs
it "finds a thing" $ do
  x <- findTheThing
  x `shouldBeJust` y
```

Doesn't type-check.

That said, I'd also never use it like that, because it brings absolutely no benefit over ``x `shouldBe` Just y``, IMO.

But I _do_ want what this function does: to get a `Just` value for later assertions but fail on `Nothing`. So I very much want this, I just don't think we should name a function "should" that is not of type `Expectation`.

So... can we just call it something else? `expectJust` or `ensureJust` or something? I think I like `expectJust` because the `Nothing` leads to an "expectation" failure.

**EDIT**: I'd also add `expectLeft` and `expectRight` since you're here ;)

chris-martin:
Added some stuff 
image

I didn't want to go too crazy with every form of expectation, instead added a general `expect` utility with the idea that you could drop in any prism.

PR #243, opened 2025-06-26
improve http exceptions and call stacks

chris-martin:


freckle-http/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-http-v0.1.0.0...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-http-v0.2.0.0...main)
+
+## [v0.2.0.0](https://github.com/freckle/freckle-app/compare/freckle-http-v0.1.0.0...freckle-http-v0.2.0.0)
+
+`MonadHttp.httpLbs` has a `HasCallStack` constraint, and instances throw `AnnotatedException`
+
+Breaking change: `httpStubbed` is now monadic rather than pure. Its errors are thrown into `IO` as
+`AnnotatedException`-wrapped `NoStubsMatched`.
```
pbrisbin, line 8: This feels like solving a problem of your own doing.

Using `error` gets you a callstack already (once `httpLbs` is fixed), which was why it was done that way. None of the reasons against `error` apply in this testing context (IMO), so there's no reason to blindly follow the rule (again, IMO). Throwing `NoStubsMatched` loses that, then wrapping it in `AnnotatedException` fixes the problem you just caused.

Am I missing anything?

pbrisbin (approved):

freckle-http/CHANGELOG.md
chris-martin: `error`'s call stack seems sort of impoverished since it's just a `String`

> None of the reasons against error apply in this testing context (IMO)

I'm not sure I agree there, especially when the error isn't coming directly from spec code but rather from within a library. Not having the errors arise from the IO you'd expect them to arise from seems like a potential source of debugging confusion?

Turning this into a custom exception holds onto all the information instead of formatting eagerly, which gives you a chance to catch it and do something else if you want. Generally wanting exception types rather than throwing strings also seems like a good rule to apply regardless of prod/test context, when it's a library that's throwing.

PR #245, opened 2025-08-08
fix(freckle-http): use TTL in set operation

pbrisbin:
The backend-agnostic cache logic encodes TTL into the cached data,
checks it on read, and ignores/deletes it if expired.

For a backend that supports expiry itself (e.g. memcached), this should
not be necessary. And, at times of heavy load, the extra `delete`
operations can cause performance problems.

To address this, we update the `set` member of `HttpCache` to accept the
TTL value too. Then, in the memcached implementation, we use it to set
with the key that TTL, meaning expired values will simply not be
retrieved by `get` at all, and no `delete`s will occur (modulo races).

We do still retain the backend-agnostic logic, since the state
implementation (for example) does not support expiring keys and so
expiration still needs to be checked externally for such backends.

Since the signature of `HttpCache.set` is changing, this must be a major
version bump, even though it is unlikely to disrupt any users in
practice.

### DIFF
--- a/freckle-app/CHANGELOG.md
+++ b/freckle-app/CHANGELOG.md
@@ -1,4 +1,8 @@
-## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-app-v1.23.3.0...main)
+## [_Unreleased_](https://github.com/freckle/freckle-app/compare/freckle-app-v1.24.0.0...main)
+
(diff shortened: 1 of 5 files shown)


pbrisbin:
> [T]he delete happens because the client was doing a GET and seeing the expired TTL encoded in the value that was retrieved rather than as an actual TTL in the cache, and now that shouldn't happen because the GET won't get anything

You are completely correct.

PR #272, opened 2026-06-11
Remove Hspec.AnnotatedException module

### LINKED ISSUE #224: Pull Freckle.App.Test.Hspec.AnnotatedException out of freckle-app
I think it could stand alone as a package, or maybe get lumped into some sort of hspec extras thing.

pbrisbin:
This module has moved to `freckle-exception`.

I considered keeping it as a pass-through, but because the name didn't
change in the new package, that would have required `PackageImports` to
do, and possibly even when it's imported by our consumers. We also chose
to remove existing re-export modules in v1.25, so we probably shouldn't
maintain one now.

While fine, it is worth mentioning that this naming issue is a symptom
of our slight laziness in extracting the various packages from
`freckle-app` without reconsidering the `freckle-` prefix or changing
the `Freckle.App` module namespace. The former is a choice about the
package's purpose, but the latter is just incorrect; module namespaces
should match their new package names: `Freckle.Exception`,
`Freckle.Http`, `Freckle.Prelude`, etc. This could be worth addressing
before same-named modules across different version of different packages
does become a problem somewhere.


chris-martin (approved):

chris-martin:
> this naming issue is a symptom
of our slight laziness in extracting the various packages from
freckle-app without reconsidering the freckle- prefix or changing
the Freckle.App module namespace

Laziness indeed, would love to go through and do a big rename, maybe bump the major vanity version numbers.

